PR #179, opened 2026-02-05
Feat/replies

pepina-dev:
## Description

This PR adds replies logic to chat screen.

⚠️  I left the scroll from the preview to the original message out of this PR cause it was getting bigger (it's more complex than I expected to do it in a way that it's not with average height assumption as the white noise app did). I'm currently working on this

**Commit by commit:**
- Adds an is replying state to use chat input hook
- Adds a new reply preview type  (extracted it cause it is used in hooks and widgets)
- Adds logic to the use chat messages to get a reply preview. For this it is necessary to load the authors metadata. Decided to get only the metadata of users who have sent a message instead of all members of group cause there could be large groups where only some of the members participate
- Adds a new reply preview widget and translation for message not found adn "you" for the case in which is a reply of a message send by me
- Adds reply support to message bubble
- Adds on reply callback to the message actions screen
- Adds replies tags support to message service
- Uses all reply logic added before in the chat screen to make replies work

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## Checklist

- [ ] Updated android .so files by running `just build-android` (only needed if rust dependencies or rust code changes)

## Summary by CodeRabbit

* **New Features**
  * Replying flow: start a reply from message actions, show inline reply preview above the input, and cancel reply.

* **UI**
  * Message bubbles display reply previews (author + excerpt); input area shows/removes preview when replying.

* **Sending**
  * Replies include reference metadata so sent messages are linked to the original.

* **Localization**
  * "Message not found" added in multiple languages.

* **Tests**
  * Expanded coverage for reply state, preview resolution, sending replies, and UI interactions.

### DIFF
--- a/lib/hooks/use_chat_messages.dart
+++ b/lib/hooks/use_chat_messages.dart
@@ -1,9 +1,9 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:whitenoise/models/reply_preview.dart';
+import 'package:whitenoise/services/user_service.dart';
 import 'package:whitenoise/src/rust/api/messages.dart';
 import 'package:whitenoise/src/rust/api/metadata.dart';
-import 'package:whitenoise/src/rust/api/users.dart' as users_api;
 
 typedef ChatMessagesResult = ({
   int messageCount,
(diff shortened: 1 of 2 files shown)


lib/hooks/use_chat_messages.dart
```diff
@@ -9,12 +12,15 @@ typedef ChatMessagesResult = ({
...
   String? latestMessageId,
   String? latestMessagePubkey,
+  ReplyPreview? Function(String? replyId) getReplyPreview,
 });
 
 ChatMessagesResult useChatMessages(String groupId) {
   final messageIds = useRef<List<String>>([]);
   final messagesById = useRef<Map<String, ChatMessage>>({});
   final indexById = useRef<Map<String, int>>({});
+  final authorsMetadataByPubkey = useRef<Map<String, FlutterMetadata>>({});
```
erskingardner: A questions on this. `useRef` doesn't trigger a rebuild when it's updated right? Won't that mean that as data is populated, no user details show up?

PR #225, opened 2026-02-11
Delete all data

### LINKED ISSUE #89: Delete all data
In the prior app we had a big button at the bottom of the app settings page or app settings screen that allowed you to delete all app data. We have a back-end method for this from the whitenoise-rs crate and FRB methods, so we should be able to just use that in the same way as we did before. And once we delete all data and clear all local preferences and settings, we should just be pushed back to the login screen.

untreu2:
## Description
Adds a "Delete All Data" button to App Settings that permanently removes all RUST DB data from the device.

Closes #89

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [x] 🧪 Tests

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a destructive "Delete all data" action in Settings with a confirmation dialog, explicit permanent-deletion warning, loading state, retryable error message, and navigation back to Home on success.

* **Localization**
  * Added localized strings for the delete-all-data flow across multiple languages (confirmation, warning, error, and button text).


lib/hooks/use_delete_all_data.dart
```diff
@@ -0,0 +1,55 @@
...
+useDeleteAllData() {
+  final state = useState(const DeleteAllDataState());
+
+  Future<void> deleteAllData() async {
+    state.value = state.value.copyWith(isDeleting: true, hasError: false);
+    try {
+      _logger.info('Deleting all application data');
+      await api.deleteAllData();
+      _logger.info('All data deleted successfully');
+      state.value = state.value.copyWith(isDeleting: false);
```
josefinalliende: I think here a `hasError: false` is missing... or am I missing something?

`state.value = state.value.copyWith(isDeleting: false, hasError: false);`

lib/widgets/wn_confirmation_bottom_sheet.dart
```diff
@@ -0,0 +1,115 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_screenutil/flutter_screenutil.dart';
+import 'package:gap/gap.dart';
+import 'package:whitenoise/l10n/l10n.dart';
+import 'package:whitenoise/theme.dart';
+import 'package:whitenoise/widgets/wn_button.dart';
+
+class WnConfirmationBottomSheet extends StatelessWidget {
```
josefinalliende: Shouldn't this look like WnSlate with the hero animation? We are using that in other screens that look like modals, like in the message actions screen.

josefinalliende (commented):
I feel that the confirmation modal should be a WnSlate instead of a sheet that looks different and above. I know there is no design yet for this, but something like this other modal that is a slate

Image

lib/hooks/use_delete_all_data.dart
```diff
@@ -0,0 +1,56 @@
...
+      _logger.info('All data deleted successfully');
+      state.value = state.value.copyWith(isDeleting: false, hasError: false);
+      return true;
+    } catch (e, stackTrace) {
+      _logger.severe('Failed to delete all data', e, stackTrace);
+      state.value = state.value.copyWith(
+        isDeleting: false,
+        hasError: true,
+      );
+      return false;
```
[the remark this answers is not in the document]
untreu2: I think this is not needed for ValueNotifier in hooks. Unlike State.setState() in stateful widgets, ValueNotifier.value = does not throw after disposal - it simply becomes a no-op. Am I right?

PR #234, opened 2026-02-12
feat(user-search): search by name with improved UX and metadata handling

### LINKED ISSUE #220: Search by name
In the user search screen, we want to be able to search users not only by npub but by name. After user serach rust methods are merged #209 , we need to include this logic on the user search screen. 

Placeholder should probably change to something that includes names too

### LINKED ISSUE #221: Metadata loading of follow list in user search screen is broken
- Something is broken now in the user search screen, I see almost all my contact lists without names or pictures 
- When I click into an avatar, I can see the metadata in next screen, so we know metadata is in db
- When I close the start chat screen and go back to the user search screen, metadata is still not shown in the user search list

Video of the issue:

https://github.com/user-attachments/assets/4e05dce1-e27a-4265-9123-42facbfb00a6

erskingardner:
## Summary

Implements search-by-name on the user search screen with several UX improvements and bug fixes.

### Added
- **Name search with batched results** — search stream updates are buffered (300ms) to reduce list jerkiness as results arrive
- **Loading indicator** — spinner in the search field suffix while name search is active
- **Metadata passthrough** — metadata is passed to the start chat screen via route `extra` so it displays instantly instead of re-fetching

### Changed
- **Redesigned user list tile** — matches the profile switcher pattern: medium avatar, middle-ellipsis npub on both title (no name) and subtitle rows, consistent vertical padding
- **Sorted follows list** — users with metadata shown first (alphabetically), then users without metadata
- **Periodic follows refresh** — re-fetches follows every 5s so metadata loaded in the background re-sorts users into the correct position
- **Truncated about text** — long about fields capped at 10 lines with ellipsis in the user profile card

### Fixed
- **UTF-16 crash** — sanitizes malformed UTF-16 strings from Nostr metadata (lone surrogates replaced with U+FFFD, plus UTF-8 round-trip validation) preventing `ArgumentError: string is not well-formed UTF-16` rendering crashes
- **Start chat metadata** — metadata from search results is forwarded to the start chat screen so the profile card isn't empty while the API re-fetches

Closes #220
Closes #221

CleanShot 2026-02-12 at 10 45 27

Video showing the features: 👇
https://share.cleanshot.com/Z7VSx5Ss

## Summary by CodeRabbit

* **New Features**
  * Search users by name with debounced, batched streaming results
  * Loading indicator shown in the search field during name searches
  * Preview metadata passed into the start-chat screen for instant display

* **Changes**
  * Redesigned user list tiles and tap behavior to surface metadata sooner
  * Follows sorted with users having metadata shown first and refreshed periodically
  * Localized search placeholder across languages
  * Long "about" text truncated to 10 lines

* **Bug Fixes**
  * Sanitize malformed UTF-16 in user metadata to prevent rendering crashes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,10 +70,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixes sign out exception and adds dedicated sign out screen with private key backup [PR #45](https://github.com/marmot-protocol/sloth/pull/45)
 - QR code color now uses theme-aware color [PR #183](https://github.com/marmot-protocol/sloth/pull/183)
 - DM avatar color inconsistency [PR #199](https://github.com/marmot-protocol/sloth/pull/199)
-<<<<<<< HEAD
 - Sanitize malformed UTF-16 in user metadata to prevent rendering crashes [PR #234](https://github.com/marmot-protocol/whitenoise/pull/234)
-=======
 - Lock app orientation to portrait mode [PR #235](https://github.com/marmot-protocol/whitenoise/pull/235)
->>>>>>> origin/master
 
 ### Security


josefinalliende (changes requested):
The follow list loading feels a lot better now that I see first results with metadata 💯 

But... I am requesting changes cause the search by name is not working for me for people I know I follow. 


Once I start searching  my phone starts getting hot, almost burning 🔥 🚒  .... and it doesn't find results by name of people that I am 100% sure I follow (like Javier haha) . They only appear once I search them by npub first, but not if I search them by name.

Steps of what happened to me:
1. I was following this npub: npub1r60m86qk2xndvmj8dr0lrnhnvc0v74q7snx460hvyy7wwvlk5nvst4sq4u
2. I searched by that account name "aleups". Nothing found (I cut the video but spent a lot more time waiting and nothing happened besides my phone getting really hot)

https://github.com/user-attachments/assets/9df72842-e5b7-4b07-9499-c7d1098ad58c

3. I searched by npub and got the result ✅ 

https://github.com/user-attachments/assets/021cbac5-3922-428f-a83f-790c6939ab75

4. Going back to the follow list then showed my friend 
 
 


I thought that maybe it was a casing issue, so I deleted the app, reinstalled and tried again but I just couldn't make it work. The I tried with other names like "Jeff" thinking that it could be a popularity issue 😆  but I couldn't find you either...

PR #250, opened 2026-02-13
feat(login): improve Amber login UX and fix disabled button opacity

### LINKED ISSUE #191: Login with Amber
Designs and description for login with Amber on Android:
https://www.figma.com/design/Y12t1SzBbrQ9Q4UTNBSoEs/02.-Application-Design?node-id=701-35050&t=DGv6x1UoVYTNI0ss-1

erskingardner:
## Summary

- Rename "Login with Signer" to **"Login with Amber"** across all 8 locales for clarity
- Disable the nsec login button when the input field is empty (previously always enabled)
- Amber button is now always outline style instead of swapping primary/outline based on input state
- **Fix disabled button opacity to match Figma**: apply `Opacity(0.25)` to the entire button container instead of per-color alpha channels — the previous approach made text nearly invisible in light mode
- Add `just run` / `just run-production` convenience commands

### Login screen (`login_screen.dart`)
- Nsec login button: always primary, disabled when empty
- Amber button: always outline
- Removed dynamic button type swapping logic

### Button widget (`wn_button.dart`)
- Disabled state now wraps the entire `FilledButton` in `Opacity(opacity: 0.25)` matching the Figma component spec
- Applies to all button types: primary, outline, ghost, overlay, destructive
- Simplified all `_build*Button` methods by removing per-color disabled alpha logic

### Tests (`login_screen_test.dart`)
- Added `await tester.pump()` after `enterText` calls (needed since login button is now disabled when empty)
- New tests for disabled/enabled state of login button
- Updated test descriptions and assertions

### Localization
- `loginWithSigner` → `loginWithAmber` in all `.arb` files + generated code

Closes #191

## Summary by CodeRabbit

* **New Features**
  * Added run commands to launch staging and production app flavors on a device.

* **Refactor**
  * Simplified login button logic and tightened disabled/enabled state handling.

* **Style**
  * Unified button visuals to use fixed semantic colors for disabled and normal states.

* **Chores**
  * Updated authentication button text across supported languages to reference Amber; added IDE ignore entries.

* **Tests**
  * Expanded login screen tests for button states, loading, errors and UI synchronization.

### DIFF
--- a/lib/widgets/wn_button.dart
+++ b/lib/widgets/wn_button.dart
@@ -100,7 +100,7 @@ class WnButton extends StatelessWidget {
     final fontSize = _getFontSize();
     final iconPadding = (size == WnButtonSize.small || size == WnButtonSize.xsmall) ? 4.w : 8.w;
 
-    Widget button = FilledButton(
+    final Widget button = FilledButton(
       onPressed: (loading || disabled) ? null : onPressed,
       style: FilledButton.styleFrom(
         padding: EdgeInsets.symmetric(vertical: verticalPadding, horizontal: horizontalPadding),


lib/l10n/generated/app_localizations.dart
```diff
@@ -1200,11 +1200,11 @@ abstract class AppLocalizations {
   /// **'Dec'**
   String get monthDecShort;
 
-  /// Button text to login using an external signer app
+  /// Button text to login using the Amber signer app (Android only)
```
josefinalliende: @erskingardner initially @dannym-arx was using the Amber word. But then we talked that (in theory) the signer flow should work with any android signer that implements NIP-55 ... so then we came to use the signer word instead of amber, for user facing texts but also in the code for naming things.

erskingardner: The design at the moment says Amber and since this only works on Android with Amber right now (we haven't tested anything else) I'm leaning towards leaving this. We can explicitly support more signers in the future once we've tested them well. What do you think?

PR #287, opened 2026-02-16
fix: remove manual scroll-to-bottom on signup screen keyboard open

### LINKED ISSUE #144: Sign up screen scrolls too far on load hiding inputs
## Summary

When the sign up screen loads, it scrolls too far, causing the input fields to be hidden from view.

## Expected Behavior

The sign up screen should load with input fields visible and accessible to the user.

## Actual Behavior

On load, the screen scrolls past the input fields, hiding them from view.

## Steps to Reproduce

1. Navigate to the sign up screen
2. Observe that inputs are not visible/scrolled out of view

## Files to Investigate

- `lib/screens/signup_screen.dart`

erskingardner:
## Summary

- Removes the `useEffect` that called `scrollController.animateTo(maxScrollExtent)` when the keyboard appeared, which was scrolling past the name input field and hiding it from view
- Removes the now-unused `scrollController` and its import
- Updates the keyboard test to verify inputs remain visible rather than validating the buggy scroll-to-bottom behavior

Flutter's `Scaffold` with `resizeToAvoidBottomInset: true` (the default) combined with built-in focus management already handles scrolling the focused field into view natively — the manual override was fighting against this.

Closes #144

## Summary by CodeRabbit

* **Refactor**
  * Streamlined keyboard behavior on the signup screen to ensure input fields remain visible when the keyboard appears.

### DIFF
--- a/lib/screens/signup_screen.dart
+++ b/lib/screens/signup_screen.dart
@@ -1,12 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_hooks/flutter_hooks.dart'
-    show
-        HookWidget,
-        useAnimationController,
-        useEffect,
-        useScrollController,
-        useState,
-        useTextEditingController;
+    show HookWidget, useAnimationController, useEffect, useState, useTextEditingController;
 import 'package:flutter_screenutil/flutter_screenutil.dart';
 import 'package:gap/gap.dart' show Gap;
 import 'package:hooks_riverpod/hooks_riverpod.dart' show HookConsumerWidget, WidgetRef;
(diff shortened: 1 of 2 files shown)


josefinalliende (approved):
Removing the scroll fixes the issue for the name field, but for the description field it gets a bit annoying cause in small screens it looks now cut. But the fix of the name field is definitely more important than the issue on the description field, so I am approving. 

Here's a video of what happens in an iphone 6:

https://github.com/user-attachments/assets/b0d72262-1dd9-4b29-abac-55b5e9eaaa8a

I think that the ideal fix would be to only scroll if the focused input that triggered the key board opening was the description, and not a scroll to bottom but scrolling a bit so that the field looks like a multiple field with more space

PR #326, opened 2026-02-20
Fix media tags to use MIP-04 v2

pepina-dev:
## Description

There is a new version for media in MIP-04 that requires a nonce tag.  This fixes the issue of media only displaying in accounts in same devices but not between different devices

## Type of Change

- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Media files can include an optional nonce and per-file scheme version; tags now emit per-file version (MIP-04 v2 when provided).
  * Media tag generation includes additional optional metadata when present (original file hash, blurhash, dimensions).
* **Chores**
  * Media tag generation refactored to build tags per-file for more accurate per-file metadata handling.

### DIFF
--- a/lib/services/message_service.dart
+++ b/lib/services/message_service.dart
@@ -30,7 +30,7 @@ class MessageService {
           )
         : <messages_api.Tag>[];
 
-    final mediaTags = await _buildMediaTags(mediaFiles);
+    final mediaTags = await _buildMediaTags(mediaFiles: mediaFiles);
     final allTags = [...replyTags, ...mediaTags];
 
     await messages_api.sendMessageToGroup(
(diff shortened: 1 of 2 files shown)


lib/services/message_service.dart
```diff
@@ -173,28 +173,39 @@ class MessageService {
...
+      'v mip04-v2',
+    ];
+    if (mediaFile.originalFileHash != null) {
+      tags.add('x ${mediaFile.originalFileHash}');
+    }
+    if (mediaFile.fileMetadata?.blurhash != null) {
+      tags.add('blurhash ${mediaFile.fileMetadata?.blurhash}');
+    }
+    if (mediaFile.fileMetadata?.dimensions != null) {
+      tags.add('dim ${mediaFile.fileMetadata?.dimensions}');
```
jgmontoya: I think we should extract `mediaFile.fileMetadata` to a variable: `final metadata = f.fileMetadata;` and use that instead of repeating the chaining

erskingardner (approved):

PR #468, opened 2026-03-07
[Fixes #445] feat: implement NIP-C7 reply threading in chat messages

### LINKED ISSUE #445: Update quote replies to follow NIP-C7
# Context

For replies we are using e tags when we should be using q tag to be interoperable with marmot ts

# Solution
- Replace the tag in the message sender service
- BUT we need to add the quoted nostr event id in the message content and then we need to parse the message content (in rust ideally) and then hide that event  id from the message content for the user

dannym-arx:
Implements NIP-C7 reply threading: replies now use a single `q` tag and prepend a `nostr:nevent1...` URI to the message content, so marmots across the colony know exactly which burrow tunnel they're replying to.

relies on: https://github.com/marmot-protocol/whitenoise-rs/pull/581

## Summary by CodeRabbit

* **New Features**
  * Replies now include a standardized nostr:nevent URI prefix and use a single consolidated reference tag.
  * Public utility added to generate nostr:nevent URIs.

* **Bug Fixes / Improvements**
  * Reply construction and tag handling streamlined for consistent formatting across message types.
  * New API error variant exposes keyring-unavailable login errors.

* **Tests**
  * Updated tests to validate URI-prefixed replies and the new tag/error behavior.

### DIFF
--- a/lib/src/rust/api/error.dart
+++ b/lib/src/rust/api/error.dart
@@ -49,6 +49,9 @@ sealed class ApiError with _$ApiError implements FrbException {
   const factory ApiError.loginInternal({
     required String message,
   }) = ApiError_LoginInternal;
+  const factory ApiError.loginKeyringUnavailable({
+    required String message,
+  }) = ApiError_LoginKeyringUnavailable;
   const factory ApiError.other({
     required String message,
   }) = ApiError_Other;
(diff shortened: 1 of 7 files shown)


josefinalliende (dismissed):
LGTM! 🎉 just resolve conflicts and you can merge!

PR #493, opened 2026-03-10
feat: cursor-based pagination for chat messages (closes #122)

### LINKED ISSUE #122: Add pagination for message fetching
## Problem

`fetch_aggregated_messages_for_group` returns *all* messages at once:
```rust
pub async fn fetch_aggregated_messages_for_group(...) -> Result<Vec<ChatMessage>, ApiError>
```

For long-running chats with thousands of messages, this will:
- Consume excessive memory on mobile devices
- Cause slow initial load times
- Eventually lead to OOM crashes

## Suggestion

Implement cursor-based pagination:
```rust
pub async fn fetch_aggregated_messages_for_group(
    pubkey: String,
    group_id: String,
    before: Option<DateTime<Utc>>,  // Fetch messages before this timestamp
    limit: Option<u32>,              // Max messages to return (default 50)
) -> Result<Vec<ChatMessage>, ApiError>
```

The Flutter side can then implement infinite scroll by requesting older messages as the user scrolls up.

## Priority
High - essential for production use with active groups

dannym-arx:
marmot

Wires up cursor-based pagination for chat messages on the Flutter side, solving the memory and OOM problem for long-running group chats. The stream's initial snapshot loads the first page, and as the marmot scrolls up toward older messages, `loadOlderMessages()` fetches the next page via the updated `fetchAggregatedMessagesForGroup` API (Rust side: `f8786caf`).

this needs a lot of testing of course...

## Summary by CodeRabbit

* **New Features**
  * Message history pagination with on-scroll "load older messages" and a top-area loading indicator.
  * Pagination entries added to the raw stream debug log.

* **Improvements**
  * More reliable initial positioning and jump-to-unread behavior.
  * Refined unseen-message indicator and read-mark timing tied to scrolling.
  * Message lists scoped per account so each account shows its own view.

* **Tests**
  * Expanded test coverage for pagination, scroll behavior, read-marking, and debug logging.


pepina-dev (changes requested):
### 👩🏼‍💻  My thoughts
Tried it in my phone and the chat seemed to be working fine BUT by debugging and adding some indicators to see when messages were fetched, I realized that pagination is not really working because the chat stream in the initial setup is returning all messages, not paginated.  

Then I confirmed by trying in master branch with chats with over 200 messages and I wasn't seeing only 50 as I expected but the 200 (which is fine for the release but means that the chat stream on the rust crate is returning just the first page when opening it).

A  concern, that  I have by reading the code, but that wasn't really able to test in a devices cause pagination is if this approach would trigger loadings/delays when users reach the top or if it would behave more like an infinite scroll.

josefinalliende:
## 🤖  My agent thoughts
 (⚠️ I tried to reproduce  the bugs mentioned by the agent and by trying to do that  Irealized that the pagination wasn't really working)... so not 100% sure if they are valid.  Maybe it makes more sense to first make the changes in rust side 🦀  for chat stream and then verify this. 

 #### 1) Stale closure bug (`groupId != requestGroupId`)
**Context:** `lib/hooks/use_chat_messages.dart`

When the user scrolls up to load older messages, the hook triggers an asynchronous fetch. If the user presses "Back" and quickly opens a different chat (changing the `groupId`) *before* the `await` completes, the newly fetched messages from the old chat will be incorrectly appended to the new chat's state, causing data corruption and severe UI bugs.

**Steps to address:**
Capture the `groupId` before the async operation begins. After the `await`, verify the component is still mounted and the `groupId` hasn't changed.
```dart
final requestGroupId = groupId;
final newMessages = await fetchOlderMessages(...);
if (!context.mounted || groupId != requestGroupId) return;
// Proceed with state update
```

###  2. `isScrollDownButtonVisible` bug
**Context:** `lib/hooks/use_chat_scroll.dart`

The logic dictating when the "Scroll to Bottom" FAB appears does not accurately track the scroll position for a `reversed: true` `ListView`. In a reversed list, `position.pixels == 0` is the absolute bottom. If the state updates indiscriminately on every scroll frame without checking if the value actually changed, it will cause severe performance degradation (unnecessary rebuilds of the entire chat tree) or trigger the button at the wrong time.

**Steps to address:**
Define a clear threshold and ensure you only update the `ValueNotifier` / state when the boolean condition actually changes to avoid unnecessary rebuilds.
```dart
final shouldShow = position.pixels > _scrollDownThreshold;
if (isScrollDownButtonVisible.value != shouldShow) {
  isScrollDownButtonVisible.value = shouldShow;
}
```

### 3. `totalMessageCount` 1-frame lag
**Context:** `lib/hooks/use_chat_messages.dart`

The `totalMessageCount` state and its corresponding `useEffect` create a double-rebuild on every new message (once from the list updating, and again when `totalMessageCount.value` changes). This 1-frame lag can cause the `ListView` item count and the underlying data array to be out of sync, resulting in brief UI judder or out-of-bounds errors during rapid scrolling.

**Steps to address:**
You can remove the `totalMessageCount` `useState` and simply return `messages.length` (or similar derived state) directly in the hook's return statement. If you strictly need the server count, group the state updates so they happen synchronously before the next frame paints.



###  4. Missing pagination trigger in debug screen
**Context:** `lib/screens/chat_raw_debug_screen.dart`

The raw debug screen (used for developer inspection of message payloads) currently loads the initial batch of messages but lacks the scroll listener to load older ones. This makes debugging older corrupted or malformed messages impossible without scrolling up in the main UI first.

**Steps to address:**
Attach the `useChatScroll` hook (or a simplified `NotificationListener<ScrollNotification>`) to the `ListView` in the debug screen and trigger the `loadMore` function when reaching the bottom of the list.

pepina-dev (changes requested):
I compared an scenario of about 500 messages in master and this branch and this is considerably faster 🎉 ... but I found some bugs  🐛 ...

I pushed a branch called `test/pepi-scenarios` with the script I was using to send messages with the CLI, so you can replicate the case I was checking. 

These are the issues I found: 
 
- The last message shown in the chat list is not matching the last message in the chat screen.... not sure why is this happening. Maybe this one is not related to this specific changes but to the fact that multiple messages were sent in the same minute, and the one saved as last for the chat summary is not matching the order of the initial message of the chat stream?

https://github.com/user-attachments/assets/0db849fc-daba-44ba-9e98-fb934483da24


- The initial scroll is not starting from the last unread message now. In this group for example I hadn't read a single message (I just accepted the invite and the messages were sent after that) so I would have expected it to start from the beggining of the conversation,  but instead it started from the end.  I understand that starting from the end makes things a lot easier ... but not sure if it is good UX. @vladimir-krstic WDYT?  

https://github.com/user-attachments/assets/0a71d870-3d49-45dc-b986-500a6bf34016

In the case of invites screen (before accepting), it looks like it was starting from the beginning but is actually not, "brave" in the video is not the real first message:

https://github.com/user-attachments/assets/128c4117-e7a2-4dde-91d3-73acbeee3b37

- The unread indicator logic now is broken. Even when I have already scrolled to the bottom, if I scroll up now the unread indicator is shown. Before, this was only visible if I had unread messages. A video of the issue:

https://github.com/user-attachments/assets/15d25439-3f89-42ee-8bdb-1098b0f64def

- When I scroll to the top, I can't see the start of the conversation,  but I can't see the first message I was expecting... this one is other position (16th instead of 1st)

https://github.com/user-attachments/assets/51a550a8-15a5-4d94-ad4a-83e017d0b9ea

pepina-dev (commented):
Overall I feel it is working better 🎉  But... I found 2 bugs 😬 

### 🐛 Issue with initial scroll in chat screen after scrolling down in invite screen
**Scenario** 
 Unaccepted invite. Receives all messages before accepting the invite. Then enter the invite screen, scroll to bottom in invite screen, accept invite, enter to chat screen

**Problem**
The initial scroll is still not working. I scrolled down in the invite screen but then when I entered the chat screen the scroll was in another place, not at the bottom as it was working on master before and I would have expected . Also (that's not in the video but it happened and checked more times) when going back to the chat list the 99+ unread count was still 99+. 

https://github.com/user-attachments/assets/cb83a229-cc1a-4f9d-afcb-3cf182d4149f


### 🐛 Missing unread indicator arrow button in chat screen of chats with lots of unread messages
**Scenario** 
Accepted invite. Receives all messages after having accepted it an going back to the chat list screen (so that all messages of the chat are unread). Then enter the chat screen


**Problem**
 When entering the chat I don't see the unread indicator arrow although I had multiple messages in the chat. Here in the video tries in two chats and in both happens the same.

https://github.com/user-attachments/assets/0df53754-c456-4736-bbe9-ddf0ae7322dc

pepina-dev (approved):
💯  It works now!! 🎉

Issue #434, opened 2026-03-03
UI Polish v0.3


vladimir-krstic:
Polish visual incorrectness in UI design, in version 0.3

[Screenshots and Screens](https://www.figma.com/design/Y12t1SzBbrQ9Q4UTNBSoEs/02.-Application-Design?node-id=1458-59297&t=7jVMRDm6a785J5uR-1) put side by side in Figma, annotated with most of the improvements that have to be made.

aleeeups:
this issue is resolved or not?

vladimir-krstic:
@aleeeups I think they are. Can't confirm right now, for some reason master won't pass Create Account for me. Looking at the PRs I would consider this done. @pepina-dev you agree?

pepina-dev:
Looking at the PR names mentioned here it looks to me that yes, that it should be fixed, although I haven't entered screen by screen checking if the figma comments were implemented completely. The only one that I don't see in the list is the profile keys screen. 

Today @vladimir-krstic shared a new UI proposal,  so personally I don't think it's worth the effort to an design that is already going to become outdated, so I rather close this issue finished or not.

PR #545, opened 2026-03-28
UI Polish: Network & Appearance Screens

untreu2:
<p align="center">
  
  
  
</p>

Related issue: #434

## Type of Change

- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## Checklist

## Summary by CodeRabbit

* **New Features**
  * "Restore default relays" action with in‑UI button and loading state
  * Dedicated Add Relay screen with category-specific flows (My, Inbox, Key Package)

* **UI/UX Improvements**
  * Replaced in-header add icons with full-width section add buttons; network list layout tweaks
  * Appearance: show system language name alongside "System"
  * Minor button styling and label text updates ("Add relay", "Relay address")

* **Localization**
  * Added relay-related strings in multiple languages

* **Tests**
  * Added/updated tests for add-relay, restore-defaults, and appearance flows; removed obsolete bottom-sheet tests


rust/src/api/accounts.rs
```diff
@@ -244,6 +244,37 @@ pub async fn account_relays(pubkey: String, relay_type: RelayType) -> Result<Vec
     Ok(relays.into_iter().map(|r| r.into()).collect())
 }
 
+const DEFAULT_RELAY_URLS: [&str; 3] = [
```
Jeezman, line 247: Noticed this is a duplicate of https://github.com/marmot-protocol/whitenoise/blob/3765196e56aefe21292fd0da7f6525efa05ce839/rust/src/api/groups.rs#L13-L17 do you want to consider extracting a shared constant? So they don't drift apart

See also https://github.com/marmot-protocol/whitenoise/blob/3765196e56aefe21292fd0da7f6525efa05ce839/rust/src/api/bug_report.rs#L14-L18

lib/hooks/use_network_relays.dart
```diff
@@ -166,10 +167,20 @@ useNetworkRelays(String pubkey) {
     }
   }
 
+  Future<void> restoreDefaultRelays() async {
+    try {
+      await accounts_api.restoreDefaultRelays(pubkey: pubkey);
+      await fetchAll();
+    } catch (e) {
+      _logger.severe('Failed to restore default relays', e);
```
Jeezman, line 175: Probably need to update the state here too the user knows something went wrong and can retry. See https://github.com/marmot-protocol/whitenoise/blob/3765196e56aefe21292fd0da7f6525efa05ce839/lib/hooks/use_network_relays.dart#L161-L164

lib/l10n/app_en.arb
```diff
@@ -1738,22 +1762,6 @@
   "@failedToStartSupportChat": {
     "description": "Error notice shown when starting a DM with support account fails"
   },
-  "fatalErrorTitle": "Oh no!",
```
Jeezman: Want to revert this?`fatal_error_screen.dart` still references https://github.com/marmot-protocol/whitenoise/blob/3765196e56aefe21292fd0da7f6525efa05ce839/lib/screens/fatal_error_screen.dart#L132 `l10n.fatalErrorDescription`, `l10n.fatalErrorCopyError`, and `l10n.fatalErrorErrorCopied`, next time flutter gen-l10n runs, the generated files will regenerate without these strings, and the fatal error screen will fail to compile

pepina-dev: Good catch! Thanks @Jeezman

vladimir-krstic:
@untreu2 @josefinalliende I can't run this branch in simulator. Each time I get stuck on splash screen error:

Error:
 
CleanShot 2026-04-06 at 17 02 52@2x

untreu2:
@vladimir-krstic can you run just generate and reset (device > erase data...) your sim?

vladimir-krstic (dismissed):

lib/widgets/wn_button.dart
```diff
@@ -18,6 +18,7 @@ class WnButton extends StatelessWidget {
     this.disabled = false,
     this.leadingIcon,
     this.trailingIcon,
+    this.iconSize,
```
pepina-dev: I don't like this approach at all. The whole point of having a design system is making the code reusable, not ending up then passing a 16.31.w icon size to a button. The previous approach to just get the icon size based in the button size made a LOT more sense to me.

lib/screens/network_screen.dart
```diff
@@ -144,86 +134,139 @@ class NetworkScreen extends HookConsumerWidget {
...
+          footer: Padding(
+            padding: EdgeInsets.fromLTRB(16.w, 24.h, 16.w, 16.h),
+            child: SizedBox(
+              width: double.infinity,
+              child: WnButton(
+                key: const Key('restore_default_relays_button'),
+                text: context.l10n.restoreDefaultRelays,
+                size: WnButtonSize.medium,
+                trailingIcon: WnIcons.reset,
+                iconSize: 16.31.w,
```
pepina-dev: My previous comment in WnButton is related to this kind if sizes. I think the icon size button being calculated based on the button size makes the code easier to read and I don't think this is a huge improvement in how the app looks. It should look fine with the default icon size of the figma component  of the application components figma

rust/src/api/accounts.rs
pepina-dev: This is going to change very soon in other PR that is in draft where I extract it to a shared constant, so I don't think it is relevant for now

lib/l10n/app_en.arb
pepina-dev: This is super weird. I checked the branch locally in commit 3765196e and the keyword is there and the fatal error screen looks fine. My guess was that the key was repeated in master but it is not. SO I don't understand this but wont block because of this

lib/hooks/use_network_relays.dart
pepina-dev: I agree that on the case of error a feedback should be shown to the user. I don't feel though it is a categorized error but a general error cause it affects all kinds of relays

pepina-dev (changes requested):
@Jeezman  commetn about handling error makes sense to me. The change in icon size button prop I does not make sense to me and I feel it is a bit against the approach of having a design system with standard button sizes to then use everywhere. The idea of the design system is that once we have the base components, we can just use the variants and it should look fine. Adding custom sizes for icons in buttons for each screen doesn't feel in line with that.

lib/widgets/wn_button.dart
untreu2: Ah yes, I had to do that because I’d used the SVGs with the wrong values, but once I corrected the size values in the SVG files, the problem sorted itself out

pepina-dev (commented):
**Blocker**
Error in restoring default relays is still swallowed.I simulated this case by adding a throw above the line `await accounts_api.restoreDefaultRelays(pubkey: pubkey)`; in use network relays hook and this is what happened: no error system notice or any kind of feedback that shows that something went wrong

https://github.com/user-attachments/assets/94efe486-b42e-4346-aaef-efbc09080fc4

**Suggestions**

Also, I noticed som tiny details about wording compared to figma, but can be handled in a separate issue in case @vladimir-krstic aggrees that the figma wording is better. I didn't find this two cases in the UI polish page so I assumed the wordings should have matched the flows figma, but maybe I'm wrong about that


- "Delete" instead of "remove"

|Figma| This branch|
|----|----|
|Image|Image|

- "Restore relays" instead of "Restore default relays"

|Figma| This branch|
|----|----|
|Image|Image|

vladimir-krstic:
@josefinalliende Different wording is fine.

pepina-dev:
> @josefinalliende Different wording is fine.

Thaks vlad! Then the only blocker would be the error swallowing @untreu2

lib/l10n/app_de.arb
```diff
@@ -173,6 +173,7 @@
   "restoreDefaultRelays": "Standard-Relays wiederherstellen",
   "restoreDefaultRelaysConfirmationTitle": "Standard-Relays wiederherstellen?",
   "restoreDefaultRelaysConfirmationMessage": "Möchtest du wirklich die Standard-Relays der App wiederherstellen? Dadurch werden deine aktuellen Relays gelöscht und ersetzt.",
+  "restoreDefaultRelaysError": "Standard-Relays konnten nicht wiederhergestellt werden. Bitte versuche es erneut.",
```
pepina-dev, line 181: TIL that Relays in german is Relais. But not a blocker cause it is in lots pf parts of this file, can be added like a separate issue of fixing Relay german translation

pepina-dev (dismissed):
✅ LGTM, now error case works good 👍🏻

pepina-dev (approved):

PR #564, opened 2026-04-13
Refactor default relays urls

pepina-dev:
## Description

#530 was getting big. We aggreed to make smaller PRs, so in that order I am dividing that one in chunks to make it easier to review. 

In this PR we extract reepated default relay urls to its own file and add a method so that they are accesible to dart (cause next PR is the offline provider which tries to connect to the default relays to check if the device has internet connection)

**What to check:**  Nothing visible to the user should have really changed, so we just need to check that this didn't  break anything. I tried bug report and network screens restore default relays and they seem to be working fine. 

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added API access to retrieve default relay URLs in both parsed (structured) and string forms.

* **Refactor**
  * Centralized default relay URL definitions into a single shared source used across APIs.

* **Tests**
  * Added unit tests verifying parsed and string default relay outputs match the shared source.

* **Chores**
  * Updated test mocks to support and simulate the new default-relay API behavior.


rust/src/api/groups.rs
```diff
@@ -9,18 +11,8 @@ use whitenoise::{
...
-];
-
 fn group_creation_relays() -> Result<Vec<RelayUrl>, ApiError> {
-    GROUP_CREATION_RELAY_URLS
-        .into_iter()
-        .map(RelayUrl::parse)
-        .collect::<Result<Vec<_>, _>>()
-        .map_err(ApiError::from)
+    default_relay_urls_parsed()
 }
```
dannym-arx: do we need this function here since it just returns the default_relay_urls_parsed() ?

dannym-arx (commented):
LGTM, just a minor comment

rust/src/api/groups.rs
pepina-dev: Oh good catch, it's not needed, will remove it

jgmontoya (dismissed):

erskingardner (dismissed):

rust/src/api/groups.rs
pepina-dev: I'll resolve this one cause it is already fixed 😅

pepina-dev: Oh no, I thought I had pushed 🤦🏻‍♀️

dannym-arx: still not pushed btw lol, but honestly this is such a minor issue where it's fine to merge it like this

pepina-dev: yes, then got stuck cause tested one more time and realized that group creation wasn't working and thought it was my fault and that I had broken that when fixing this comment. But it wasn't the case, that bug is already in master.

pepina-dev: Left more details in https://github.com/marmot-protocol/whitenoise-rs/issues/749

jgmontoya (approved):

PR #573, opened 2026-04-15
Add block/unblock hook

untreu2:
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [x] 🧪 Tests

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added block/unblock functionality with a toggle to block or unblock users; shows per-action and overall loading states and surface errors with a clear-error option.

* **Tests**
  * Added comprehensive tests verifying loading states, block/unblock/toggle behavior, error handling, refresh behavior, and cleanup.

* **Chores**
  * Updated changelog entry for the new block/unblock hook.

### DIFF
--- a/lib/hooks/use_block_actions.dart
+++ b/lib/hooks/use_block_actions.dart
@@ -26,25 +26,33 @@ BlockActionsState useBlockActions({
   final error = useState<String?>(null);
 
   useEffect(() {
+    if (userPubkey == null) {
+      isLoading.value = false;
+      return null;
+    }
+
+    var cancelled = false;
+    isLoading.value = true;
+
     Future<void> fetchIsBlocked() async {
-      if (userPubkey == null) return;
-      isLoading.value = true;
       try {
         final result = await mute_list_api.isUserBlocked(
           accountPubkey: accountPubkey,
           targetPubkey: userPubkey,
         );
+        if (cancelled) return;
         isBlocked.value = result;
       } catch (e) {
+        if (cancelled) return;
         _logger.severe('Failed to fetch block status: $e');
         isBlocked.value = false;
       } finally {
-        isLoading.value = false;
+        if (!cancelled) isLoading.value = false;
       }
     }
 
     fetchIsBlocked();
-    return null;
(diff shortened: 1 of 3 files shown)


lib/hooks/use_block_actions.dart
```diff
@@ -0,0 +1,120 @@
...
+
+typedef BlockActionsState = ({
+  bool isBlocked,
+  bool isLoading,
+  bool isActionLoading,
+  String? error,
+  void Function() clearError,
+  Future<void> Function() block,
+  Future<void> Function() unblock,
+  Future<void> Function() toggleBlock,
```
pepina-dev, line 13: do block and unblock need to be public in the state? It looks to me that with `isBlocked `and `toggleBlock` exposing those methods should't be necessary... or am I missing something?

lib/hooks/use_block_actions.dart
```diff
@@ -0,0 +1,120 @@
...
+      _logger.severe('Failed to unblock user: $e');
+      error.value = 'Failed to unblock user';
+      rethrow;
+    } finally {
+      isActionLoading.value = false;
+    }
+  }
+
+  Future<void> toggleBlock() async {
+    if (userPubkey == null) return;
```
pepina-dev: if the state isLoading then the isBlocked value could be null but just because it hasn't been loaded yet. So I think in isLoading case it should return too...

test/hooks/use_block_actions_test.dart
```diff
@@ -0,0 +1,445 @@
...
+        expect(getState().error, isNull);
+
+        await expectLater(getState().unblock, throwsException);
+        await tester.pump();
+
+        expect(getState().error, 'Failed to unblock user');
+      });
+    });
+
+    group('toggleBlock action', () {
```
pepina-dev: There is a missing test case for the case I mentioned above, when the state is loading

pepina-dev (approved):
✅ LGTM

PR #590, opened 2026-04-24
feat(rust): add group_required_proposals method

jgmontoya:
## Description

Exposes group capabilities from the rust API be able to hide/show features

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## Checklist

- [ ] Run `just precommit` to ensure that formatting and linting are correct
- [ ] Updated the `CHANGELOG.md` file with your changes (if they affect the user experience)

## Summary by CodeRabbit

* **New Features**
  * Added a new API to retrieve required proposals for groups, exposing group proposal information with support for self-remove and additional proposal types.

* **Chores**
  * Updated dependencies to latest compatible versions.

PR #571, opened 2026-04-14
Leave group hook

pepina-dev:
## Description

Adds leave group to rust bridge, adds small hook to leave group and adds the translations needed for the leave action from chat list

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## Checklist

- [ ] Run `just precommit` to ensure that formatting and linting are correct
- [ ] Updated the `CHANGELOG.md` file with your changes (if they affect the user experience)

## Summary by CodeRabbit

* **New Features**
  * Users can now leave groups via a "Leave group" action with a confirmation dialog, warning about chat behavior, and error feedback on failure.

* **Localization**
  * Added translations for the leave-group flow in English, German, Spanish, French, Italian, Portuguese, Russian, and Turkish.

* **Tests**
  * Added tests validating loading state, error handling, duplicate-call deduplication, and safe unmount during leave.

PR #600, opened 2026-04-28
Use group capabilities in leave group hook

pepina-dev:
## Description

First step to leave group was adding a hook to call the leave method in #571 . BUT we then realized with @jgmontoya that not all groups would be able to allow members to leave without causing problems. This is because old versions of the app don't have a way to process a leave event. So, we need to show leave button only for groups  that have the MLS capability to leave.  In #590 jgmontoya exposed mls group capabilities and now we can use them when to show the leave button.

**This PR:**
- Adds a feature flag provider to be able to have this feature in master but disabled (in order to not block a release)
- Adds translations that will be used in step with the UI from chat list
- Adds selfRemoved to the chat summary type. This is needed to then detect when we where removed or we left from the group, to show the correct wording
- Adjusts the leave group hook to decide if it canLeave or not, in order to only show the button to leave if it is possible

**What to test:**
With or without the feature flag enabled, no UI changes should be noticeable, so the only thing that needs to be tested is that the app keeps working as usual compared to master.

To run the app with the feature flag, just add -`-dart-define LEAVE_GROUP_ENABLED=true`

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## Checklist

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/600">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Feature-flag controlled ability to leave groups, with UI gating and capability checks.

* **Improvements**
  * Updated post-leave UX and confirmations; chat lists now reflect voluntary exits so items update correctly.
  * Shorter "Leave" action label and new post-leave messages added across DE/EN/ES/FR/IT/PT/RU/TR.

* **Tests**
  * Expanded test coverage for leave flows, gating, and chat-list behavior; test fixtures adjusted accordingly.


test/hooks/use_leave_group_test.dart
```diff
@@ -45,38 +66,221 @@ void main() {
...
+    mockApi.shouldThrowOnProposals = false;
     mockApi.wasLeaveCalled = false;
     mockApi.leaveCallCount = 0;
+    mockApi.proposalsCallCount = 0;
     mockApi.passedPubkey = null;
     mockApi.passedGroupId = null;
     mockApi.leaveCompleter = null;
+    mockApi.requiredProposals = [RequiredProposal.selfRemove];
+    mockApi.proposalsCompleter = null;
+  });
```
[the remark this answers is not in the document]
pepina-dev: This kind of comments are annoying when tests are working fine

lib/constants/feature_flags.dart
```diff
@@ -0,0 +1,7 @@
+enum FeatureFlag { leaveGroup }
+
+const _featureFlagValues = {
+  FeatureFlag.leaveGroup: bool.fromEnvironment('LEAVE_GROUP_ENABLED'),
```
cypherpinkdev, line 4: This doesn't make easy to have mutliple feature flags at a time if they have to be called ith --dart-define BUT I guess it is on purpose  that we don't want to have many feature flags at the same time?

lib/constants/feature_flags.dart
```diff
@@ -0,0 +1,7 @@
...
+};
+
+bool featureFlagValue(FeatureFlag flag) => _featureFlagValues[flag] ?? false;
```
cypherpinkdev, line 7: We should add though an issue to remember to delete this feature flag once it is defaulted to true and activated in release

lib/l10n/app_en.arb
```diff
@@ -1807,7 +1807,11 @@
   "@removedFromGroup": {
     "description": "Notice title shown when the user has been removed from the group"
   },
-  "removedFromGroupDescription": "You can still view saved messages, but you can't send or receive new ones. You can archive or delete the chat anytime.",
+  "youLeftThisGroup": "You left this group",
+  "@youLeftThisGroup": {
+    "description": "Notice title shown in chat screen when the user voluntarily left the group"
+  },
+  "removedFromGroupDescription": "You can still view saved messages, but you can't send or receive new ones. You can archive the chat anytime.",
```
cypherpinkdev, line 1814: The design says delete. I guess this is because we don't have delete option yet right?

cypherpinkdev (changes requested):
Checks are failing with a `useMounted` is deprecated warning

lib/constants/feature_flags.dart
pepina-dev: yes, that is on purpose

lib/l10n/app_en.arb
pepina-dev: Yes, it's temporary in order to avoid confusing users

lib/constants/feature_flags.dart
pepina-dev: https://github.com/marmot-protocol/whitenoise/issues/602

cypherpinkdev (approved):
LGTM but lets wait for another review

dannym-arx (approved):

PR #601, opened 2026-04-28
Add Fastlane release scaffolding

erskingardner:
## Summary
- Add Fastlane as a thin wrapper around the existing Flutter and Rust release scripts
- Add a release version validator plus tests for tag/version matching
- Add `just` entrypoints for staging and production release builds
- Update Android staging to use `dev.ipf.whitenoise.staging`
- Document current app IDs and add Zap Store APK matching for production releases

## Testing
- Added shell tests for release version validation and release automation config
- Verified Fastlane Ruby files parse cleanly
- Verified `just release-doctor` and tag validation pass
- Verified `flutter build apk --flavor staging --debug --no-pub` produces an APK with the expected staging package ID
- Verified Zap Store YAML still parses

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/601">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive release process guide covering version checks, build targets, artifact staging, and a manual release runbook.

* **Chores**
  * Added release automation tooling and runnable tasks to build and stage Android/iOS artifacts; included Fastlane and task wrappers.
  * Updated VCS ignores to exclude local credential and build artifacts.
  * Made staging Android application ID explicit.

* **Tests**
  * Added automated tests to validate release scripts, version/tag handling, and release-task presence.

* **Bug Fixes**
  * Narrowed artifact selection rules to target arm64 APKs.


pepina-dev (approved):
LGTM! ✅  I tried running ` just test-release-scripts` and` just release-doctor` and both worked

scripts/validate_release_version.sh
```diff
@@ -0,0 +1,148 @@
...
+    sed -E 's/^[[:space:]]*version:[[:space:]]*//' |
+    sed -E 's/[[:space:]]*#.*$//' |
+    sed -E 's/^[[:space:]]*["'\'']?//; s/["'\'']?[[:space:]]*$//'
+)"
+if [[ "$FULL_VERSION" != *+* ]]; then
+  echo "Release version must include a build number, e.g. 2026.4.28+23" >&2
+  exit 1
+fi
+
+VERSION_NAME="${FULL_VERSION%%+*}"
```
dannym-arx, line 95: I think we should enforce YYYY.MM.DD versioning here

android/app/build.gradle.kts
```diff
@@ -140,7 +140,7 @@ android {
     productFlavors {
         create("staging") {
             dimension = "environment"
-            applicationIdSuffix = ".staging"
+            applicationId = "dev.ipf.whitenoise.staging"
```
dannym-arx: I'm not sure about switching the applicationId

pepina-dev: I think that for staging there is no problem. That is just for us for testing purposes. Buuut now that I'm writing it, it makes a lot of sense not to change it cause otherwise we won't be able to tests cases in staging that could be happening in the real app because we swiped the data in the staging app

pepina-dev (changes requested):
I regret my approve, cause I aggree with danny on not changing the staging bundle id ... unless we decide to change the production one too. Otherwise we will have cases in real apps that we won't be able to replicate in staging app cause staging apps will start with wiped data.

pepina-dev:
> I regret my approve, cause I aggree with danny on not changing the staging bundle id ... unless we decide to change the production one too. Otherwise we will have cases in real apps that we won't be able to replicate in staging app cause staging apps will start with wiped data.

I thought changing the staging bundle id wasn't a big issue because it was only for testing purposes, but I forgot that it could make our QA and testing worse.

scripts/validate_release_version.sh
erskingardner: we can't. it's really annoying but flutter's pubspec breaks with leading zeros in the version because it treats them as semver values.

android/app/build.gradle.kts
erskingardner: yup moved back to the old one for now.

pepina-dev (approved):
GTG!

PR #617, opened 2026-05-04
[codex] Bump whitenoise-rs master and refresh Android assets

erskingardner:
## Summary

- Rebase onto current `master` and resolve the Cargo conflicts by keeping the latest upstream `whitenoise-rs` master pin: `b30c9ae34f0b078077ae056ea314f135a3fe7fb7`.
- Refresh `Cargo.lock` for the rebased dependency graph.
- Keep the release version at `2026.5.4+23` and keep Widgetbook's path dependency lock entry aligned.
- Include the DM avatar loading-state fix/test and the updated padded Android staging icon assets.
- Use one shared conservative `isGroupChat` flag in `ChatScreen` so loading DM profile state is not treated as a group chat.

## Validation

- `git ls-remote https://github.com/marmot-protocol/whitenoise-rs refs/heads/master`
- `cargo update -p whitenoise --manifest-path rust/Cargo.toml`
- `just regenerate`
- `flutter test test/screens/chat_screen_test.dart`
- `just precommit`
- `just build-android-quiet`
- `just build-release-android` produced `build/releases/v2026.5.4+23/`
- Verified production arm64 release artifact metadata: `versionName=2026.5.4`, `versionCode=2023`, package `org.parres.whitenoise`

## Review finding status

- Skipped the request to change `pubspec.yaml` back to `2026.3.23+29`: stale/invalid. The intended release version is now `2026.5.4+23`, and the release tooling confirms the matching output path `build/releases/v2026.5.4+23/`.
- Addressed the `ChatScreen` group-chat consistency finding: valid. The shared flag uses `chatProfile.data?.isDm == false`, so null/loading state remains non-group and does not reintroduce the DM avatar flash.

## Notes

- PR branch was rebased onto `origin/master` and force-pushed with lease.
- The release-channel signing continuity still needs to be checked against the last published GH/Zapstore APK before publishing.
- The local worktree still has an unrelated unstaged deletion of `reviews/PR-517-start-chat-flow-improvements.md`; it is intentionally not part of this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed chat type detection logic to properly determine group chats versus direct messages, improving consistency in message bubble rendering and display options.

* **Chores**
  * Updated app version to 2026.5.4+23
  * Updated package dependencies


pepina-dev (commented):
I've been testing this branch and comparing it to master and I am facing the problem that we talked about in the morning:

📱🅰️ : iPhone 2026.3.23+22
📱🅱️: iPhone in this branch
📱🆑 : Android in this branch


**Testing 📱🅰️ with 📱🅱️**
- ❌  I can't start a chat (DM or group) from A to B.  I get the warning of "Invite to White Noise" which means that the rust side is returning that it didn't find a key package. When I check in developer settings of B, I see both kinds of key packages published, it's not the case that the account does not have a legacy keypackage. 

- ✅  What works is in the other direction (starting a chat from B to A), both DMs and groups. **This was failing in master but not always, in the morning it wasn't and now it was working perfect for me. 

**Testing 📱🅱️ with 📱🆑**
- Everything seems to be working perfect accross devices of the same version


🐛  Something that worries me is that now, even after deleting the app in📱🅱️, I can't run master after having run this branch, when I switch back to master and try to run the app I get this error: 
```

ApiError.whitenoise(message: Database error: Database encryption key error: Keyring error: Stored key has invalid length (expected 32 bytes): Invalid encryption key length: expected 32 bytes, got 62 bytes)

#0      SimpleDecoder.decode (package:flutter_rust_bridge/src/codec/base.dart:32)
#1      SseCodec._decode (package:flutter_rust_bridge/src/codec/sse.dart:45)
#2      SseCodec.decodeObject (package:flutter_rust_bridge/src/codec/sse.dart:35)
<asynchronous suspension>
#3      initializeAppContainer (package:whitenoise/main.dart:74)
<asynchronous suspension>
#4      main (package:whitenoise/main.dart:55)
<asynchronous suspension>
```

I tried installing deps again, running just regenerate, but nothing worked. In 📱🅰️, where I didn't run this branch, running master works, so is something with this upgrade that seems to have left things in a weird state... but maybe it is expected? 

### 🤔Conclusions?
I'm worried about the first error (❌) cause I tested this in master and it was working. On the other direction (✅ ) although it seems to be working well, it is also working well for me in master now, so I'm not really sure if something improved in this direction or not.

mubarakcoded (approved):
LGTM!

erskingardner:
@pepina-dev I've done a bunch of testing and... to recap on your issues:

### Old clients not being able to invite new clients 

I actually don't think we have an issue here. I just ran the test by starting clean, checking what KPs both clients had on relays manually with `nak` and then doing invites in both directions. Both worked just fine. My suspicion is that our previous test was done on a phone that had had the app installed for a while and potentially had an older/invalid key package published. Or potentially that relays just hadn't quite had time to index the new client's key package? I'd be curious if you can make it happen repeatedly though.

### Database keys error

This is completely expected. In https://github.com/marmot-protocol/mdk/pull/270 I fixed a bug where we were putting raw bytes into the secret store. Some keyring stores only support valid utf-8 strings. That PR migrates the raw bytes to utf-8 base64 encoded strings. 

When you delete the app, it doesn't clear the secret store on the device. It probably should, but currently doesn't. So when you reinstalled the older version of the app, it saw a value so it didn't try and migrate anything but the older version of the code is looking for a 32-byte key, it has no knowledge of the fix or the new format. I can probably clean this up a little too but it's not fundamentally an error in the process users will go through since they can't downgrade an app.

pepina-dev (commented):
@erskingardner But the problem is starting a chat from the old release with this branch, so the key package that is not found by the old app is the key package published by this version, that's my big concern. I tested  multiple times and it happened all the time, it wasn't flaky, but will test again.

pepina-dev:
> @erskingardner But the problem is starting a chat from the old release with this branch, so the key package that is not found by the old app is the key package published by this version, that's my big concern. I tested multiple times and it happened all the time, it wasn't flaky, but will test again.

This is SO WEIRD. Now it just works. In both directions.

pepina-dev (approved):
LGTM

PR #619, opened 2026-05-05
Save to gallery

### LINKED ISSUE #421: Save images in gallery
# Context

Images are downloaded in the chat, but are not saved in gallery. We need an option to save images in gallery.

pepina-dev:
## Description

Closes #421 Allows to save in gallery images and videos from the media modal with long press or with the download button. 

⚠️ 🎨 At first I was showing the success feedback as a system notice in the media modal, but as it used space above the image, it reduced the space available and forced the image to change size, which caused a visual "jump" of the image. So I added a check in the same download button cause the success, which feels to me more subtle. For the error case I kept the system notice. 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## How I Tested This

- Checked that download does not turn back to first index 
- Check that images and videos are actually saved in gallery
- Check that when video is playing, the download button hides

## Screenshots

https://github.com/user-attachments/assets/e669d277-e0fa-4938-a673-3b106ec8fbb0

## Checklist

- [ ] `just precommit` passes
- [ ] Related issue linked (`Closes #NNN`)
- [X] `CHANGELOG.md` updated (if user-visible change)
- [ ] Screenshots added (for UI changes)

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/619">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Save-to-gallery for images/videos with UI download control, overlay support on media players/previews, and inline system notices for failures and brief “saved” confirmation.

* **Localization**
  * Added save-to-gallery error messages in English, German, Spanish, French, Italian, Portuguese, Russian, and Turkish.

* **Chores**
  * Updated platform manifest/permission entries for gallery access (Android/iOS) and added gallery-handling dependency.

* **Tests**
  * Expanded test coverage for save flows, overlays, and aspect-ratio parsing.


ios/Runner/Info.plist
```diff
@@ -71,5 +71,7 @@
 		<false/>
 		<key>NSCameraUsageDescription</key>
 		<string>Camera access is required to scan QR codes</string>
+		<key>NSPhotoLibraryAddUsageDescription</key>
+		<string>White Noise needs access to save images to your photo library.</string>
```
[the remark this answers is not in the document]
pepina-dev: removed this cause it was untranslated

test/hooks/use_save_to_gallery_test.dart
```diff
@@ -0,0 +1,412 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:gal/gal.dart';
+import 'package:gal/src/gal_platform_interface.dart';
```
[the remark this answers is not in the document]
pepina-dev: Tests works, so I'll leave it as it is for now

lib/widgets/media_modal.dart
```diff
@@ -150,13 +169,35 @@ class _MediaContent extends StatelessWidget {
...
   });
 
   @override
   Widget build(BuildContext context) {
+    final l10n = context.l10n;
+    final currentFile = mediaFiles[currentIndex];
+    final mediaDownload = useMediaDownload(mediaFile: currentFile);
+    final localPath = mediaDownload.status == MediaDownloadStatus.success
+        ? mediaDownload.localPath
+        : null;
```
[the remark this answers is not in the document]
pepina-dev: media modal always receives an array of media files... so this  guard makes no sense IMO

lib/hooks/use_save_to_gallery.dart
```diff
@@ -0,0 +1,82 @@
...
+}) {
+  final status = useState(SaveToGalleryStatus.idle);
+  final error = useState<SaveToGalleryError?>(null);
+  final savedRecently = useState(false);
+
+  useEffect(() {
+    savedRecently.value = false;
+    error.value = null;
+    return null;
+  }, [localPath]);
```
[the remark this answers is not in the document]
pepina-dev: local path shouldn't change in mid save... I think this is a weird case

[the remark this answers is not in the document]
pepina-dev: Lets see if it really causes issues fist

erskingardner:
Why do we need an icon at the top right. Shouldn't we just be able to do a long-press on the image and then get a sub-menu? We definitely shouldn't change this now but I'm curious about why we didn't do that to start. @vladimir-krstic

vladimir-krstic:
@erskingardner it's a common affordance in all other messaging apps. So just to be in line with expected experience.  They have a share modal button instead which we will have as well eventually. Now this makes me wonder, why don't we simply have a Share button. Ofc, don't do it now, but @pepina-dev is this more complicated in any way?

ios/Runner/Info.plist
```diff
@@ -69,7 +69,5 @@
 		<true/>
 		<key>UIStatusBarHidden</key>
 		<false/>
-		<key>NSCameraUsageDescription</key>
```
erskingardner, line 72: My agent flagged this. Strange that this didn't show up as problems in the build though...

> The `gal` package's own documentation states that `NSPhotoLibraryAddUsageDescription` is **required** for iOS, and `NSPhotoLibraryUsageDescription` is required for iOS < 14. Neither key is added by this PR. Without `NSPhotoLibraryAddUsageDescription`, the app will either crash or silently fail when users attempt to save media on iOS, and it will be rejected by App Store review.
> 
> Additionally, the PR removes `NSCameraUsageDescription`, but `mobile_scanner` (already in the project) still needs camera access for QR scanning. That key is used in all the `*.lproj/InfoPlist.strings` localized files that this PR does not touch — removing the root plist entry while leaving the localized strings is inconsistent and will cause a warning/error on iOS.

android/app/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,9 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_DATA_SYNC"/>
     <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS"/>
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
```
erskingardner: > The `gal` README specifies the permission should be declared as:
> 
> ```xml
> <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"
>                  android:maxSdkVersion="29" />
> ```
> 
> Without `maxSdkVersion="29"`, the app will declare this permission on API 30+ where it is no longer required and where declaring it without the cap is unnecessary (and potentially raises a Play Store review flag). The `maxSdkVersion` attribute is what `gal`'s own example uses and is the accepted pattern.

macos/Flutter/GeneratedPluginRegistrant.swift
```diff
@@ -25,6 +26,7 @@ func RegisterGeneratedPlugins(registry: FlutterPluginRegistry) {
   FileSelectorPlugin.register(with: registry.registrar(forPlugin: "FileSelectorPlugin"))
   FlutterLocalNotificationsPlugin.register(with: registry.registrar(forPlugin: "FlutterLocalNotificationsPlugin"))
   FlutterSecureStorageDarwinPlugin.register(with: registry.registrar(forPlugin: "FlutterSecureStorageDarwinPlugin"))
+  GalPlugin.register(with: registry.registrar(forPlugin: "GalPlugin"))
```
erskingardner, line 29: We should probably add this so that we're not confused in the future with MacOS

> The `gal` README states that `macos/Runner/Info.plist` also requires `NSPhotoLibraryAddUsageDescription`. This PR registers the macOS `GalPlugin` but does not add the macOS permission string. If macOS is a target, this will cause failures.

lib/widgets/media_modal.dart
erskingardner, line 1: > Per the project's architecture, hooks (ephemeral state) belong in the screen-level widget or a dedicated hook widget, and data flows down via parameters. `_MediaContent` is a private, screen-scoped widget — adding hooks to it is not wrong per se, but it now has both `currentIndex` passed as a prop *and* it calls `useMediaDownload` for `mediaFiles[currentIndex]` internally. If `currentIndex` changes, the hook re-runs with a new `MediaFile`, which is fine — but if this widget is ever reused or its scope changes, the implicit dependency on `currentIndex` being in sync with external state will be non-obvious. At a minimum the `onError` callback passed to `useSaveToGallery` closes over `l10n` which is fine, but capturing `onSaveError` (a prop) in the closure means the hook sees the callback from the most recent build, which is the correct Flutter behaviour. No action required beyond awareness.

lib/widgets/media_modal.dart
```diff
@@ -169,16 +210,60 @@ class _MediaContent extends StatelessWidget {
               onPageChanged: onPageChanged,
               itemBuilder: (_, index) {
                 final mediaFile = mediaFiles[index];
+                final aspectRatio = getAspectRatioFromDimensions(
+                  mediaFile.fileMetadata?.dimensions,
+                );
+
+                final downloadButton = WnIconButton(
```
erskingardner, line 217: > `downloadButton` is instantiated inside `itemBuilder`, which is called for each page. Because it uses a constant key (`const Key('media_modal_download_button')`), Flutter will reuse the element across pages, but the construction still happens per-call. It would be cleaner to hoist this out of the `itemBuilder` since it does not vary by page (all pages share the same `saveToGallery` state). This would also make the intent clearer.

pepina-dev:
> @erskingardner it's a common affordance in all other messaging apps. So just to be in line with expected experience. They have a share modal button instead which we will have as well eventually. Now this makes me wonder, why don't we simply have a Share button. Ofc, don't do it now, but @pepina-dev is this more complicated in any way?

Yes we can do it. In that case I would need a different dependency. Taking a quick look, something like share plus  package could be useful

vladimir-krstic:
@pepina-dev cool! You don't have to do this now unless it's like dumb simple and won't break anything else.

ios/Runner/Info.plist
pepina-dev: This is weird, it says it is always required, but I tried in my iPhone and it just works without it... 🤔

macos/Flutter/GeneratedPluginRegistrant.swift
pepina-dev: Good catch

lib/widgets/media_modal.dart
pepina-dev: I have a refactor to how media modal handles media downloads in other branch, but didn't add it to this PR to get this shipped for the release. So I'll reevaluate this after the refactor to see if it still applies

pepina-dev:
> @pepina-dev cool! You don't have to do this now unless it's like dumb simple and won't break anything else.

 I think I'd prefer to handle that in another PR to get this merged for tomorrow (🤞🏻)  release.

pepina-dev:
https://github.com/marmot-protocol/whitenoise/issues/631

ios/Runner/Info.plist
pepina-dev: ✅ brought it back and added translations for this descriptions.

android/app/src/main/AndroidManifest.xml
pepina-dev: Added 👍🏻  thanks for the catch

macos/Flutter/GeneratedPluginRegistrant.swift
pepina-dev: Added ✅

lib/widgets/media_modal.dart
pepina-dev: I decided to use a different key per page, using the index of the page for it

erskingardner (approved):

PR #623, opened 2026-05-06
Filter blocked users across Flutter messaging surfaces

erskingardner:
## Summary
- Hide invites, chat previews, message timelines, and search results from blocked users
- Suppress blocked senders from new chat updates and notifications
- Refresh block state from the Rust mute list so the UI stays in sync after block and unblock actions
- Exclude `widgetbook/**` from the root analyzer so Widgetbook is checked in its own package context

## Testing
- Added and updated unit and widget tests for chat list, chat screen, invite flow, message search, chat list search, and notification handling
- Ran `just generate`
- Ran `flutter analyze --fatal-infos` for the root app and for `widgetbook`
- Ran `just precommit` successfully

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/623">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * App-wide blocking: block/unblock now hides blocked users across chats, messages, searches, invites, and notifications; blocked invites are auto-redirected/hidden and unread/preview counts are sanitized.

* **Chores**
  * Analyzer configuration updated and minor dependency revision applied.

* **Tests**
  * Extensive tests added/updated covering blocking, filtering, invites, notifications, chat summary, message/search filtering, and related UI flows.


dannym-arx (dismissed):

lib/hooks/use_chat_list.dart
```diff
@@ -79,10 +118,50 @@ ChatListResult useChatList(String pubkey, {bool archived = false}) {
...
+    groupType: chat.groupType,
+    createdAt: chat.createdAt,
+    groupImagePath: chat.groupImagePath,
+    groupImageUrl: chat.groupImageUrl,
+    pendingConfirmation: chat.pendingConfirmation,
+    welcomerPubkey: chat.welcomerPubkey,
+    archivedAt: chat.archivedAt,
+    removedAt: chat.removedAt,
+    selfRemoved: chat.selfRemoved,
+    unreadCount: chat.unreadCount,
```
mubarakcoded, line 167: I tested this on a real device using this branch: Inside the chat, messages are correctly hidden 🚀

We have an issue with the chat list, It still shows unread badge counts for chats where every unread message is from a blocked user. The badge keeps inflating as new messages come in from the blocked user.
We could zero unreadCount here alongside the lastMessage redaction. This covers the common case, but misses a group where the latest unread is from a blocked user and earlier ones aren't - we'd hide legitimate unreads.
I think the cleaner long-term fix is to filter blocked authors out of the unread count on the Rust side, so the number arriving here is already correct. WDYT?

erskingardner:
@mubarakcoded great catch. I'm going to fix the badge count in whitenoise-rs like you recommended. that fix isn't a blocker for this PR

lib/screens/chat_screen.dart
```diff
@@ -86,6 +87,13 @@ class ChatScreen extends HookConsumerWidget {
     final pubkey = ref.watch(accountPubkeyProvider);
     final isOffline = ref.watch(offlineProvider).value ?? false;
     final debugLog = ref.read(messageDebugLogProvider.notifier);
+    final blockActionsRefreshKey = useState(0);
+    final blockedPubkeysState = useBlockedPubkeys(pubkey);
+    final chatList = useChatListWithBlockedPubkeys(
```
pepina-dev: No no no. I realized that emir had already introduced a fetch of the whole chat list in the chat screen just to check if I was removed from the chat or not. I asked @jgmontoya to make a method to avoid that and get just the required info instead of the whole chat list.

pepina-dev (changes requested):
I really dislike this approach of filtering everywhere in the flutter when the rust side already knows that the user is blocked.  It feels it makes  everything more complicated for flutter in lots of screens and if our goal is to keep this UI layer as simple as possible, for me this logic that is more complicated, belongs to the rust side. 

Some things here feel like an overkill. For example, we don't need to fetch all the blocked pubkeys through the chat list stream in a DM chat screen  to then decide if that single user is blocked. It should be easier, no chat list stream needed at all.

**From product perspective**

I don't like this behaviour either: 
- I should be able to see the past messages with a blocked user, and only not seeing the new messages since I blocked.  But here, since I block, all the past messages are hidden too
- If I unblock, I shouldn't see the messages that the user sent me while was blocked. But here, if I unblock a user, I can see all the messages that he sent me while I had blocked him. This for me is an important part of the blocking feature, simply "loosing" the connection with that person while block with no way to know if that person wrote to me or not. For exes, stalkers, etc this behaviour helps to reduce anxiety a lot, you simply can't know what they wrote while blocked even if at a moment of weakness you unblock them back. 

I tried for example blocking Javier in whatsapp and signal and it both and it behaves that way. After blocking him, I can enter the chat and see the previous messages before I blocked him, but I can't receive new messages. He sent messages to me while blocked and even after unblocking him, I can't see the messages he sent in the meantime.


**Strong opinion**
I'd rather release without blocking users logic (by simply removing the buttons or hiding them with a feature flag) and build it right, moving this filtering behavior to rust side and making it behave as blocking users works in whatsapp (which I think is the one that has the best UX for blocking)

pepina-dev:
> I really dislike this approach of filtering everywhere in the flutter when the rust side already knows that the user is blocked. It feels it makes everything more complicated for flutter in lots of screens and if our goal is to keep this UI layer as simple as possible, for me this logic that is more complicated, belongs to the rust side.
> 
> Some things here feel like an overkill. For example, we don't need to fetch all the blocked pubkeys through the chat list stream in a DM chat screen to then decide if that single user is blocked. It should be easier, no chat list stream needed at all.
> 
> **From product perspective**
> 
> I don't like this behaviour either:
> 
>     * I should be able to see the past messages with a blocked user, and only not seeing the new messages since I blocked.  But here, since I block, all the past messages are hidden too
> 
>     * If I unblock, I shouldn't see the messages that the user sent me while was blocked. But here, if I unblock a user, I can see all the messages that he sent me while I had blocked him. This for me is an important part of the blocking feature, simply "loosing" the connection with that person while block with no way to know if that person wrote to me or not. For exes, stalkers, etc this behaviour helps to reduce anxiety a lot, you simply can't know what they wrote while blocked even if at a moment of weakness you unblock them back.
> 
> 
> I tried for example blocking Javier in whatsapp and signal and it both and it behaves that way. After blocking him, I can enter the chat and see the previous messages before I blocked him, but I can't receive new messages. He sent messages to me while blocked and even after unblocking him, I can't see the messages he sent in the meantime.
> 
> **Strong opinion** I'd rather release without blocking users logic (by simply removing the buttons or hiding them with a feature flag) and build it right, moving this filtering behavior to rust side and making it behave as blocking users works in whatsapp (which I think is the one that has the best UX for blocking)

Signal has though a better UX though for replies of messages sent during the blocking period, that something Signal does better than whatsapp regarding to blocking

erskingardner:
@pepina-dev I hear you. There are some real constraints in our world that those centralized services don't have to deal with though...

The biggest of which is that we need to process enough MLS to decrypt/authenticate and distinguish application messages from commits/proposals. Commits/proposals from blocked members still need to be processed to keep the group state coherent unless the product decision is “blocking also removes/forks group membership,” which is a much larger protocol decision.

We _maybe_ could delete them after processing but that feels brittle to me. 

If we really want blocking that will work the way that you describe it's almost certainly going to need to be a Marmot protocol level concern. 

One thing that we might do (not perfect, I'll admit up front) would be to ship this now, and test it with some real users in Oslo later this month and get real user feedback. If it's clear that your way is much better, then we should take it seriously and do the work to make it protocol level so all clients will handle it the same. What do you think?

pepina-dev:
@erskingardner
- It makes sense that at MLS level we need to keep processing things form blocked users, I'm not against that
- What I think is that visibility of messages could be stored in rust to avoid this filtering in flutter side logic. (I think it should be more efficient cause the visibility of a message for an account could be stored in db and no need to recalculate visibility every time we enter a chat). 
- If nostr mute lists are not enough for block feature (because to emulate block behavior form other apps we need to know when it was blocked), then a MIP makes sense to me. 

😖 I'd personally prefer not to ship like this.... but you're the boss 😎 and I agree with you that it could help to get user feedback on this feature could help 

🙏 If we do ship this PR in this release, I would still like to have some things changed before merge:

1. Not using the chat list in chat screen. I know it was already used by the removed warning, but that was unnecessary IMO and that's why I asked @jgmontoya to make a method in rust crate to be able to fix that by getting a single chat summary (that I think it was opened yesterday). What I don't like about this PR is that it is using the chat list in the chat screen and refreshing it when I wanted it to be removed from that screen before this.
2. Renaming this feature to "Mute". I think it would be more accurate wording and avoid confusions with how block works in other apps. 

🙏🙏  **One petition more**
 If we release this, I would still like to work after this is merged with @jgmontoya to move as much as possible the filtering logic from flutter to rust side, hopefully for the next release before Oslo.  I think this hooks for chat list and chat messages are getting too complex... Could we agree on that in exchange for merging this for this release?

pepina-dev:
>   2. Renaming this feature to "Mute". I think it would be more accurate wording and avoid confusions with how block works in other apps.



I disagree with myself about 2. Mute in chats means another thing, more related to notifications and sounds.

pepina-dev (dismissed):
I'll approve this one in order to not block the release. I tried to make the refactor for the point 1 I mentioned earlier BUT when testing #633  I realized that I can't  get rid of the chat list stream subscription in each chat screen as I would have liked without more changes in the  rust side to emit an event for block/unblock.

pepina-dev:
> I'll approve this one in order to not block the release. I tried to make the refactor for the point 1 I mentioned earlier BUT when testing #633 I realized that I can't get rid of the chat list stream subscription in each chat screen as I would have liked without more changes in the rust side to emit an event for block/unblock.

I was mistaken, I can get rid, I was just confused. So will let you know as soon as #633 is ready for review

pepina-dev:
Now https://github.com/marmot-protocol/whitenoise/pull/633 which is pointing to this branch is ready for review

dannym-arx (changes requested):
I agree with pepi that we can move a lot of this logic to the rust side, even if we don't move it to whitenoise-rs yet. we can just put the filtering in the rust code here?

dannym-arx (approved):

mubarakcoded (approved):

PR #635, opened 2026-05-08
fix(android): external signer notification recovery

erskingardner:
## Upstream Dependency

Resolved: marmot-protocol/whitenoise-rs#811 has merged, and this PR points at the merged `whitenoise-rs` master revision `49e5f20c43fd72f1d07ea3675350754650955371`.

## Summary

- Fix Android background notifications for multi-account external-signer users after reboot / headless foreground-service startup.
- Reconcile external signer callbacks from both the UI isolate and foreground-task isolate so every persisted external-signer account has a live callback before notification handling begins.
- Register external signer callbacks before ensuring Rust relay subscriptions and before attaching the headless notification stream.
- Register app-local Android plugins with the headless foreground-task Flutter engine so the Android signer method channel is available outside `MainActivity`.
- Keep headless notification subscriptions from requesting notification permission when no Activity is attached.
- Keep the foreground service configured with the `dataSync` service type without restarting an already-running service on every cold start.
- Make external signer reconciliation best-effort for non-active accounts while preserving fast-fail behavior for the activating account.

## Review Follow-up

- Renamed the lifecycle helper from `ExternalSignerRegistrationService` to `ExternalSignerCallbackRegistry` to make the split from `AndroidSignerService` clearer.
- Added Android platform guards in both the callback registry and `AndroidSignerService.registerExternalSigner`.
- Removed the short-pubkey logging helper that implied invalid pubkeys were expected.
- Removed the `@visibleForTesting` entry point from `NotificationTaskHandler`; tests now exercise retry behavior through task start/repeat/coordination events.
- Removed the process-local foreground-service restart flag that would stop/start a running service on every cold start.
- Collapsed foreground-task subscription tracking to the subscription source of truth, and clear `NotificationSubscription.isRunning` when the stream closes.
- Left the local-account concern as code-is: reconciliation filters `AccountType.external_` before registration, and mixed local/external account tests cover that path.

## Root Cause

The boot/reboot foreground task was alive and subscribed to the Rust notification stream, but it did not have external-signer callbacks registered in that headless Dart isolate. Incoming events for Amber-backed accounts could arrive and be persisted, but Rust could not fully process/decrypt them into notification updates until the main app isolate opened and `authProvider` registered the signer callbacks.

The desired runtime state is now reconciled in both isolates: every signed-in external-signer account has a registered signer callback in the current process/isolate before notification handling begins.

## Manual Verification

Verified on a Pixel 7 Pro with two Amber-backed accounts and a second Android sender device:

- Rebooted the Amber device.
- Waited for the silent `White Noise connected to relays` foreground notification.
- Sent messages to both Amber-backed accounts before opening White Noise.
- Both Android notifications appeared while the app remained unopened.
- Logs showed the expected order: system foreground task start, external signer callback registration for both accounts, Rust relay subscription ensure, notification stream subscription, received notification updates, displayed notifications.

## Testing

- `flutter test --reporter compact test/services/android_signer_service_test.dart test/services/external_signer_callback_registry_test.dart test/services/foreground_service_test.dart test/providers/auth_provider_test.dart`
- `just test-flutter-quiet`
- `just precommit`
- Earlier Android validation: `just when-apk` plus install/manual reboot testing on physical devices.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/635">
  <picture>
    Open in Stage
  </picture>
</a>

## Release Notes

* **New Features**
  * External signer callback registration is now automatically triggered when the app resumes.

* **Improvements**
  * Enhanced external signer account reconciliation across login, logout, and profile-switching flows.
  * Notification permission requests are now optional and configurable.
  * Improved Android foreground service lifecycle handling and coordination.

* **Tests**
  * Expanded test coverage for external signer registration, app lifecycle, and notification handling.

### DIFF
--- a/lib/providers/auth_provider.dart
+++ b/lib/providers/auth_provider.dart
@@ -3,7 +3,7 @@ import 'package:flutter_secure_storage/flutter_secure_storage.dart';
 import 'package:logging/logging.dart';
 import 'package:whitenoise/providers/is_adding_account_provider.dart';
 import 'package:whitenoise/services/android_signer_service.dart';
-import 'package:whitenoise/services/external_signer_registration_service.dart';
+import 'package:whitenoise/services/external_signer_callback_registry.dart';
 import 'package:whitenoise/src/rust/api/accounts.dart' as accounts_api;
 import 'package:whitenoise/src/rust/api/error.dart';
 
@@ -15,7 +15,7 @@ final secureStorageProvider = Provider<FlutterSecureStorage>(
 );
 
 class AuthNotifier extends AsyncNotifier<String?> {
-  final _externalSignerRegistrationService = ExternalSignerRegistrationService();
+  final _externalSignerCallbackRegistry = ExternalSignerCallbackRegistry();
 
   @override
   Future<String?> build() async {
@@ -168,7 +168,7 @@ class AuthNotifier extends AsyncNotifier<String?> {
       final remainingAccounts = await accounts_api.getAccounts();
       final otherAccounts = remainingAccounts.where((a) => a.pubkey != pubkey).toList();
       final nextAccount = otherAccounts.isEmpty ? null : otherAccounts.first;
-      await _externalSignerRegistrationService.reconcile(
+      await _externalSignerCallbackRegistry.reconcile(
         remainingAccounts,
         requiredPubkeys: nextAccount == null ? const {} : {nextAccount.pubkey},
       );
@@ -244,7 +244,7 @@ class AuthNotifier extends AsyncNotifier<String?> {
     Iterable<accounts_api.Account> knownAccounts = const [],
     Set<String> requiredPubkeys = const {},
   }) async {
-    await _externalSignerRegistrationService.ensureRegistered(
+    await _externalSignerCallbackRegistry.ensureRegistered(
       knownAccounts: knownAccounts,
       requiredPubkeys: requiredPubkeys,
     );
--- a/lib/providers/auth_provider.dart
+++ b/lib/providers/auth_provider.dart
@@ -16,6 +16,8 @@ final secureStorageProvider = Provider<FlutterSecureStorage>(
 
 class AuthNotifier extends AsyncNotifier<String?> {
   final _externalSignerCallbackRegistry = ExternalSignerCallbackRegistry();
+  final _registeredExternalSignerPubkeys = <String>{};
+  final _externalSignerRegistrationFutures = <String, Future<void>>{};
 
   @override
   Future<String?> build() async {
@@ -170,6 +172,8 @@ class AuthNotifier extends AsyncNotifier<String?> {
       final nextAccount = otherAccounts.isEmpty ? null : otherAccounts.first;
       await _externalSignerCallbackRegistry.reconcile(
         remainingAccounts,
+        registeredExternalSignerPubkeys: _registeredExternalSignerPubkeys,
+        externalSignerRegistrationFutures: _externalSignerRegistrationFutures,
         requiredPubkeys: nextAccount == null ? const {} : {nextAccount.pubkey},
       );
       if (nextAccount != null) {
@@ -237,7 +241,7 @@ class AuthNotifier extends AsyncNotifier<String?> {
     await storage.write(key: _storageKey, value: account.pubkey);
     state = AsyncData(account.pubkey);
     ref.read(isAddingAccountProvider.notifier).set(false);
-    _logger.info('Login completed for ${account.pubkey}');
+    _logger.info('Login completed for ${account.accountType.name} account');
   }
 
   Future<void> _ensureExternalSignersRegistered({
@@ -245,6 +249,8 @@ class AuthNotifier extends AsyncNotifier<String?> {
     Set<String> requiredPubkeys = const {},
   }) async {
     await _externalSignerCallbackRegistry.ensureRegistered(
+      registeredExternalSignerPubkeys: _registeredExternalSignerPubkeys,
+      externalSignerRegistrationFutures: _externalSignerRegistrationFutures,
       knownAccounts: knownAccounts,
       requiredPubkeys: requiredPubkeys,
     );
--- a/lib/providers/auth_provider.dart
+++ b/lib/providers/auth_provider.dart
@@ -3,6 +3,7 @@ import 'package:flutter_secure_storage/flutter_secure_storage.dart';
 import 'package:logging/logging.dart';
 import 'package:whitenoise/providers/is_adding_account_provider.dart';
 import 'package:whitenoise/services/android_signer_service.dart';
+import 'package:whitenoise/services/external_signer_registration_service.dart';
 import 'package:whitenoise/src/rust/api/accounts.dart' as accounts_api;
 import 'package:whitenoise/src/rust/api/error.dart';
 
@@ -14,8 +15,7 @@ final secureStorageProvider = Provider<FlutterSecureStorage>(
 );
 
 class AuthNotifier extends AsyncNotifier<String?> {
-  final Set<String> _registeredExternalSignerPubkeys = {};
-  final Map<String, Future<void>> _externalSignerRegistrationFutures = {};
+  final _externalSignerRegistrationService = ExternalSignerRegistrationService();
 
   @override
   Future<String?> build() async {
@@ -37,6 +37,7 @@ class AuthNotifier extends AsyncNotifier<String?> {
 
     await _ensureExternalSignersRegistered(
       knownAccounts: activeAccount == null ? const [] : [activeAccount],
+      requiredPubkeys: activeAccount == null ? const {} : {activeAccount.pubkey},
     );
 
     if (activePubkey == null || activePubkey.isEmpty) return null;
@@ -161,15 +162,17 @@ class AuthNotifier extends AsyncNotifier<String?> {
     final storage = ref.read(secureStorageProvider);
 
     await accounts_api.logout(pubkey: pubkey);
-    _registeredExternalSignerPubkeys.remove(pubkey);
     await storage.delete(key: _storageKey);
 
     try {
       final remainingAccounts = await accounts_api.getAccounts();
-      await _reconcileExternalSigners(remainingAccounts);
       final otherAccounts = remainingAccounts.where((a) => a.pubkey != pubkey).toList();
-      if (otherAccounts.isNotEmpty) {
-        final nextAccount = otherAccounts.first;
+      final nextAccount = otherAccounts.isEmpty ? null : otherAccounts.first;
+      await _externalSignerRegistrationService.reconcile(
+        remainingAccounts,
+        requiredPubkeys: nextAccount == null ? const {} : {nextAccount.pubkey},
+      );
+      if (nextAccount != null) {
         await storage.write(key: _storageKey, value: nextAccount.pubkey);
         state = AsyncData(nextAccount.pubkey);
         _logger.info('Logout successful - switched to another account');
@@ -197,7 +200,10 @@ class AuthNotifier extends AsyncNotifier<String?> {
     final storage = ref.read(secureStorageProvider);
     try {
       final account = await accounts_api.getAccount(pubkey: pubkey);
-      await _ensureExternalSignersRegistered(knownAccounts: [account]);
+      await _ensureExternalSignersRegistered(
+        knownAccounts: [account],
+        requiredPubkeys: {account.pubkey},
+      );
       await storage.write(key: _storageKey, value: pubkey);
       state = AsyncData(pubkey);
       _logger.info('Profile switched successfully');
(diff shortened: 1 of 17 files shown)


lib/services/external_signer_registration_service.dart
```diff
@@ -0,0 +1,114 @@
+import 'package:logging/logging.dart';
+import 'package:whitenoise/services/android_signer_service.dart';
+import 'package:whitenoise/src/rust/api/accounts.dart' as accounts_api;
+
+final _logger = Logger('ExternalSignerRegistration');
+
+class ExternalSignerRegistrationService {
```
cypherpinkdev: I feel it is confusing to have 2 services now related to external signers (`AndroidSignerService ` and this new one` ExternalSignerRegistration`).  I think that if all logic is related to an External signer it would be simpler to have all this methods grouped under a single service. WDYT?

lib/services/external_signer_registration_service.dart
```diff
@@ -0,0 +1,114 @@
...
+  ExternalSignerRegistrationService({
+    Future<List<accounts_api.Account>> Function()? getAccounts,
+    Future<void> Function(String pubkey)? registerExternalSigner,
+  }) : _getAccounts = getAccounts ?? accounts_api.getAccounts,
+       _registerExternalSigner =
+           registerExternalSigner ?? const AndroidSignerService().registerExternalSigner;
```
cypherpinkdev: This is for me super confusing to read. So the service in charge of registering an external signer need to receive the method to register it? So the other service is in part in charge too of registering a signer. That makes me think that this 2 should just live together in a single service as I mentioned above.

lib/services/external_signer_registration_service.dart
```diff
@@ -0,0 +1,114 @@
...
+    _externalSignerRegistrationFutures[pubkey] = registration;
+    try {
+      await registration;
+    } finally {
+      _externalSignerRegistrationFutures.remove(pubkey);
+    }
+  }
+
+  String _pubkeyPreview(String pubkey) {
+    if (pubkey.length <= 8) return pubkey;
```
cypherpinkdev: When this case could happen? that seems to me like an invalid pubkey...

lib/providers/auth_provider.dart
```diff
@@ -14,24 +15,33 @@ final secureStorageProvider = Provider<FlutterSecureStorage>(
...
-    } catch (e) {
-      if (e is ApiError_Whitenoise && e.message.contains('Account not found')) {
-        await storage.delete(key: _storageKey);
-        return null;
+    accounts_api.Account? activeAccount;
+    var activePubkey = pubkey;
+
+    if (pubkey != null && pubkey.isNotEmpty) {
+      try {
+        activeAccount = await accounts_api.getAccount(pubkey: pubkey);
```
cypherpinkdev, line 31: Why the condition to check if an account was external or not was removed?

`if (account.accountType == accounts_api.AccountType.external_) `

 This might be problematic for accounts created with login with nsec in devices where amber is present and used but for other accounts. Unless I'm missing something.

cypherpinkdev: I see that this condition was moved inside of  complete login , but `_ensureExternalSignersRegistered `is still called here

lib/services/foreground_service.dart
```diff
@@ -112,17 +145,82 @@ class _NotificationTaskHandler extends TaskHandler {
   @override
   void onNotificationDismissed() {}
 
+  @visibleForTesting
```
cypherpinkdev: I personally try to avoid using visible for testing approach. 

> 🤖 Avoid @visibleForTesting — tests that reach internal methods break on harmless refactors and can hit full coverage while the public interface is broken. Use it only when the behavior is impossible to exercise otherwise.

lib/services/external_signer_registration_service.dart
cypherpinkdev: My agent dissagrees with me a bit in the argument for having them joined in one service, but aggrees in the naming confusion. 

> 🤖 The new `ExternalSignerRegistrationService` is a lifecycle orchestrator layered on top of it. The split is technically legitimate:
> 
> - `AndroidSignerService` = NIP-55 wire protocol (sign event, encrypt, decrypt, `registerExternalSigner`)
> - `ExternalSignerRegistrationService` = decides *when* and *for which accounts* to call `registerExternalSigner`, with deduplication and multi-account coordination
> 
> However, the naming causes real confusion. `ExternalSignerRegistrationService` sounds like it could be absorbed into `AndroidSignerService`, especially since `AndroidSignerService.registerExternalSigner` already exists and the new service's only non-trivial responsibility is deduplication and reconciliation logic. The conceptual split is defensible but the naming makes it look like unnecessary proliferation.

lib/providers/auth_provider.dart
cypherpinkdev: My agent dissagrees with me and does not think this is a real issue. Let's wait for pepi's QA review though

`### 🤖 Is the non-external signer case handled correctly?

**Yes, correctly handled at two defensive layers:**

1. In `reconcile()`: filters to `account.accountType == AccountType.external_` before building the map — local accounts never enter the registration path.
2. In `_registerExternalSignerIfMissing()`: explicit early return `if (account.accountType != accounts_api.AccountType.external_) return`.

A local account that happens to have Amber installed will never trigger external signer registration. The test `'does not re-register for local account'` (existing) and `'re-registers inactive external signer accounts'` (new, mixed local+external scenario) both verify this correctly.`

lib/services/external_signer_registration_service.dart
cypherpinkdev: My agent raised a point here too and I think it makes sense: this service by its name seems to be  more generic than just for android signers, but then it defaults to AndroidSignerService which is confusing without having a platform check when used... ) 

This is the original comment my agent left:

> **🤖 [must]** The service hardcodes `AndroidSignerService().registerExternalSigner` as its default with no platform guard. `AndroidSignerService` is Android-only. On iOS, `AccountType.external_` should never exist (iOS users can't log in via Amber), but this is an **implicit invariant, not an enforced one**. If a corrupted or cross-platform data store somehow produced an `external_` account on iOS, this service would attempt to invoke the Android plugin.
> 
> `AndroidSignerService.registerExternalSigner` itself also does NOT check `Platform.isAndroid` before calling `signer_api.registerExternalSigner` — so the platform check would cascade all the way to a Rust FFI call with unregistered Android callbacks.

lib/services/foreground_service.dart
cypherpinkdev, line 1: @pepina-dev please try to replicate this case to see if this is a real issue (was raised by my agent)
> 🤖 **[must]** `_startedWithCurrentOptions` is a process-local boolean initialized to `false`. This means on **every cold start** where a foreground service is already running, the code will unconditionally stop and restart the service to "refresh options." The intent is to pick up `ForegroundServiceTypes.dataSync`, but since `dataSync` is now always passed on start, the one-time migration is over — yet the stop/restart happens on every cold start forever.
> 
> This causes a notification icon flicker on every app launch for existing users who already have the service running. It's not a one-time migration cost; it's a permanent regression.

lib/services/foreground_service.dart
```diff
@@ -58,46 +59,78 @@ void _startCallback() {
...
+  final Future<void> Function()? _ensureSubscriptionsOverride;
+  final Future<bool> Function()? _startSubscriptionOverride;
+  final Future<void> Function()? _stopSubscriptionOverride;
+
   NotificationSubscription? _subscription;
+  ExternalSignerRegistrationService? _externalSignerRegistrationService;
   Locale _cachedLocale = const Locale('en');
   bool _bootstrapped = false;
+  bool _shouldOwnSubscription = false;
+  bool _subscriptionRunning = false;
```
cypherpinkdev: > 🤖 **[suggestion]** `_subscriptionRunning` (new field) and `_subscription != null` (pre-existing pattern in `_startSubscription`) track the same concept through different code paths. `ensureSubscriptionRunning()` checks `_subscriptionRunning`, while `_startSubscription()` returns early based on `_subscription != null`. If these ever diverge (e.g., the stream closes unexpectedly while `_subscriptionRunning` stays `true`), the `onRepeatEvent` health check becomes a no-op and the subscription never recovers.
> 
> **How to address:** Collapse to one source of truth. `_subscriptionRunning` could be derived from `_subscription?.isRunning ?? false` (since `NotificationSubscription` already exposes `isRunning`), or remove `_subscriptionRunning` entirely and let `_startSubscription`'s `_subscription != null` check be the idempotency guard.

cypherpinkdev (commented):
Left some comments about the code, haven't done QA yet

pepina-dev (changes requested):
Full trust mode seems to be working well 🎉. BUT for the other amber cases, I still see issues (marked with ❌ or 🐛  below). 

🤔 I'm wondering if we could just restrict to use amber in full trust mode.... and have it just as a way of having nsec isolated  in a single signer app but not with the purpose of  allow manually every event to be signed 

Here's a detail of the things I tested grouped by kind of permissions and how it went. 

❌ = doesn't work here and in master
🐛 =it was working better in master and doesn't in this PR 
✅ = didn't work in master and it is fixed by this PR 

## Manual grant permission mode
⚠️ In this case other accounts on the device get broken too even if they aren't using amber and no invites arrive at all. I tried this multiple times. 

Account N°1 (with amber)
- Logged in with amber (with **manual grant permission**)
- Stayed in chat list in that account
- Sent an invite and message from other device
- ❌ didn't get the notification
- ❌ didn't get the invite in chat list

Account N°2 (without amber)
- Then I Added another account by signing up (so  **without** amber) 
- Stayed in chat list in that account
- Sent an invite and message to that account 
- ❌  Didn't get the notification
- ❌ Didn't get the invite in chat list


## Approve basic actions mode 
⚠️ I didn't get to tests this with multiple accounts, I can't make it work with just one. Tried multiple times. I checked and the app does have notifications permissions

- Logged in with amber (with approved basic actions)
- Stayed in chat list in that account
- Sent an invite and message from other device
- 🐛 didn't get the notification (in master in chat list amber was opened and then after approved, the notification arrived. Here it never arrived. )
- did  get the invite in chat list

- Put the app in background
- Sent another  invite and message from other device
-  ❌  didn't  get the notification
- did  get the invite in chat list

- Closed  the app
- Sent another  invite and message from other device
- ❌ didn't get the notification

- Reboot phone
- Wait for whitenoise connected to relays
- Send message from other device
- ❌ didn't get the notification


## Full trust mode  

Account N°1 (with amber)
- Logged in with amber (with approved basic actions)
- Stayed in chat list in that account
- Sent an invite and message from other device
- get the notification
- get the invite in chat list


- Put the app in background
- Sent another  invite and message from other device
- get the notification


- Closed  the app
- Sent another  invite and message from other device
- ✅ get the notification

- Reboot phone
-  Wait for whitenoise connected to relays
- Send message from other device
-  ✅  Get the notification


Account  N°2 (without amber)
- Then I Added another account by signing up (so  **without** amber) 
- Stayed in chat list in that account
- Sent an invite and message to that account 
-  get the notification
-  get the invite in chat list

- Switched to the other account account
- Sent an invite and message to that account 
-  get the notification
- get the invite in chat list


- Switched again to the other  account
- Put the app in background
- Sent an invite and message to that account 
-  ✅ get the notification
- get the invite in chat list


- Switched again to the other  account
- Close the app
- Sent an invite and message to that account 
-   ✅ get the notification
-  get the invite in chat list

- Open the app
- Reboot phone
- Wait for whitenoise connected to relays
- Send message from other device to accounts 1 and 2
- ✅  Get the notifications for both


Account N°3 (with amber)
- Then I added a third account with amber login
- Stayed in chat list in that account
- Sent an invite and message to that account 
-  get the notification
- get the invite in chat list

- Switched to the other account 
- Sent an invite and message to that account 
-  get the notification
- get the invite in chat list


- Switched again to the other  account
- Put the app in background
- Sent an invite and message to that account 
-  ✅  get the notification
-  get the invite


- Switched again to the other  account
- Close the app
- Sent an invite and message to that account 
-  ✅  get the notification

- Reboot phone
- Wait for whitenoise connected to relays
- Send message from other device to accounts 1 , 2 and 3
- ✅  Get the notifications for all of them

pepina-dev (commented):
Oops I requested changes BUT I meant to comment cause most of the issues I found are marked with ❌  (which means they are also in master) and the one with 🐛  (introduced here) it's little compared with the things fixed with ✅

erskingardner:
@pepina-dev I think we should accept the current amber limitations, this is a clear benefit for getting multi-account and after-reboot notifications working better.

pepina-dev:
> @pepina-dev I think we should accept the current amber limitations, this is a clear benefit for getting multi-account and after-reboot notifications working better.

Yes, agree on that.  Re reviewing this for a last time.

lib/services/foreground_service.dart
pepina-dev: I don't remember seeing that flicker... so it doesn't seem a blocker to me if this happens, wasn't noticeable to my eye

lib/services/foreground_service.dart
```diff
@@ -28,22 +30,21 @@ const _kPendingNotificationTapKey = 'pending_notification_tap';
...
+  Logger.root.onRecord.listen((record) {
+    if (record.loggerName != 'ForegroundService' &&
+        record.loggerName != 'ExternalSignerCallbackRegistry' &&
+        record.loggerName != 'NotificationSubscription') {
+      return;
+    }
+    final buf = StringBuffer('[${record.level.name}] ${record.loggerName}: ${record.message}');
+    if (record.error != null) buf.write(' | error: ${record.error}');
+    if (record.stackTrace != null) buf.write(' | stackTrace: ${record.stackTrace}');
+    debugPrint(buf.toString());
```
pepina-dev: Please remove debug print. If you want that log in the release app, log it with logger. `debugPrint `still logs in release mode (it's name is confusing)

erskingardner: Replaced with developer.log (we can't use the normal logger here because we're outside the main isolate.

pepina-dev (approved):

PR #638, opened 2026-05-08
Leave group from chatlist for non-last admins 

pepina-dev:
## Description

This PR adds self demote for non last admin in chat list leave flow. For last admin, leaves leave button disabled with a message explaining why leave option is disabled. Also adds an explanation when user can't leave because of group capabilities (this case is for groups where there are users that don't support leave yet). 

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## How I Tested This
To try this feature, run the app with --dart-define LEAVE_GROUP_ENABLED=true feature flag

**Case where I'm not the last admin**
- Create a group
- Make the other user admin
- From chat list, long press the group, select leave option
- Check that I am able to leave
- Click leave
- Check that it works

**Case where I am the last admin**
- Create a group
- From chat list, long press the group, select leave option
- Check that I am not able to leave
- Make the other user admin
- From chat list, long press the group, select leave option
- Check that I am able to leave
- Click leave
- Check that it works

**Case of group with no capabilities**
 - Send a group invite from another device in an old version of the app (march release version is enough)
 - Accept the invite
 - From chat list, long press the group, select leave option
 - Check that I am not able to leave and that there is an explanation according to this case. 

## Screenshots

Not last admin case:

https://github.com/user-attachments/assets/34ec1dab-905a-4373-9ea8-5e3047d12170

Last admin case

https://github.com/user-attachments/assets/22f9c4fc-330d-448a-a130-2c71979173fe

Group without leave capabilities

https://github.com/user-attachments/assets/9f2c25ce-0041-41e2-9d1b-7e75f07926b1

## Checklist

- [ ] `just precommit` passes
- [ ] Related issue linked (`Closes #NNN`)
- [x] `CHANGELOG.md` updated (if user-visible change)
- [x] Screenshots added (for UI changes)

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/638">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Leave groups directly from the chat list; admins follow an admin-specific leave flow.
* **Behavior Changes**
  * Users who are the only admin see a warning and must assign another admin before leaving.
  * Groups created with older app versions that don’t support leaving show an informational warning.
* **Localization**
  * Added leave-group warning translations in multiple languages.
* **Tests**
  * Expanded tests covering visibility states and admin/non-admin leave flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/638)


lib/hooks/use_leave_group.dart
```diff
@@ -34,43 +40,82 @@ LeaveGroupState useLeaveGroup({
...
+                adminPubkeys.value = admins;
+              }
+            })
+            .catchError((Object e, StackTrace st) {
+              _logger.severe('Failed to fetch group admins', e, st);
+              if (!isStale) {
+                adminPubkeys.value = [];
+              }
+            }),
+      ]);
```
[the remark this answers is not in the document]
pepina-dev: I think this is ok. If the user is too fast on clicking leave while admin state is loading, it will fail and show an error. But the default case when the group has nostr capabilities for leave is to show the leave button. So I prefer to leave it with the default behaviour while admin pubkeys are loading and then, disabling the button if it is needed

lib/widgets/chat_list_tile.dart
```diff
@@ -198,6 +199,24 @@ Future<void> _runAction({
...
 }
 
+List<ChatListAction> _buildLeaveWarningActions({
+  required AppLocalizations l10n,
+}) => [
+  ChatListAction(
+    id: 'close_leave_warning',
+    label: l10n.cancel,
+    icon: WnIcons.closeLarge,
+    onTap: () async {},
```
erskingardner: this might be the intended behavior but agent flagged:

> `onTap` is non-null, so `autoDismiss: true` (default) fires, which dismisses the overlay. But the warning step has a `onBack` callback that returns to the initial menu — that back handler is never wired to the cancel button. The user pressing "Cancel" in the warning state gets the full overlay dismissed, rather than returning to the initial menu. This seems intentional given the screenshots, but it's asymmetric: the back-arrow returns to the initial menu, while "Cancel" fully dismisses. If that's the intended design, a comment would help; if not, the `onTap` should call `menuController?.updateState(buildInitialActions())` instead.

lib/widgets/chat_list_menu.dart
erskingardner, line 157: Out of the diff range but another agent finding: 

> The `currentSystemNotice` widget is rendered inside the card but its height is not estimated in `menuContentHeight`. This height estimate is used to compute `maxTop`, which positions the menu so it stays on screen. When `systemNotice` is shown (fetch error case), the menu card will be taller than estimated and may overflow off the bottom of the screen on small devices. The existing `estimatedMiddleContentHeight` of `80.h` follows the same pattern — an `estimatedSystemNoticeHeight` guard should be added alongside it.

erskingardner (dismissed):
Only a few minor things I left comments on, I'm not even sure are real issues.

lib/widgets/chat_list_tile.dart
pepina-dev: I think it makes sense to go back instead of dismiss. Now it goes back 👍🏻  thanks!

lib/widgets/chat_list_menu.dart
pepina-dev: Thanks, I couldn't replicate on my smallest phone, but added the fix anyways. Makes sense 👍🏻  thanks for the catch

cypherpinkdev (approved):
Let's merge this, it just works and jeff had previously approved

PR #641, opened 2026-05-08
fix: preserve chat list state across screen remounts (#483)

### LINKED ISSUE #483: [Bug]: Loading icon when going back to chat list
When going from a chat back to the chat list screen it shows a loading spinner, it should instead show the list instantly and update it lazily in the background. 
Same with the users profile picture, it shows a placeholder for a couple frames.

dannym-arx:
marmot

Closes #483.

Returning to the chat list flashed a loading spinner because go_router's `goNamed` disposes `ChatListScreen`, and on pop the `useChatList` hook re-subscribed to the stream with no cached data. So every back-navigation looked like a cold start.

Cache the chat-list map (per pubkey + archived) and the blocked-pubkeys set (per pubkey) at module scope so they outlive widget disposal. `isLoading` now gates on whether an initial snapshot was ever received, not on the current stream's connection state. A remount renders the cached marmots immediately and updates lazily as new stream events arrive.

Regression test in `test/hooks/use_chat_list_test.dart` reproduces the bug (fails on old code, passes on the fix). Screen-level test in `test/screens/chat_list_screen_test.dart` locks in the user-flow behavior. Test-side cache reset is wired into `MockWnApi.reset()` so individual test files keep their existing setUp.

`just precommit` passes. Coverage 99.36%.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/641">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat list preserves cached chats when returning from a conversation, eliminating spinner flicker and unnecessary loading.
  * Blocked-contacts and chat-list caching prevent redundant fetches, improving navigation speed and reducing data usage.

* **Tests**
  * Added/updated tests to cover cached chat behavior and remount scenarios; test teardown now clears related cached state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/641)

### DIFF
--- a/lib/hooks/use_blocked_pubkeys.dart
+++ b/lib/hooks/use_blocked_pubkeys.dart
@@ -35,6 +35,11 @@ BlockedPubkeysState useBlockedPubkeys(String accountPubkey, {int refreshKey = 0}
 
   useEffect(() {
     var cancelled = false;
+    // Only surface a loading state on the first fetch. Subsequent refreshes
+    // (refreshKey or manualRefreshKey changes after the cache is populated)
+    // intentionally keep `isLoading` false so callers like `_useChatList`
+    // keep showing the existing chat list while we re-fetch in the
+    // background, rather than blanking out and showing a spinner.
     if (!cache.hasLoaded) isLoading.value = true;
 
     Future<void> fetchBlockedPubkeys() async {
--- a/lib/hooks/use_blocked_pubkeys.dart
+++ b/lib/hooks/use_blocked_pubkeys.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:logging/logging.dart';
@@ -12,21 +13,38 @@ typedef BlockedPubkeysState = ({
   VoidCallback refresh,
 });
 
+class _BlockedPubkeysCacheEntry {
+  Set<String> blockedPubkeys = <String>{};
+  bool hasLoaded = false;
+}
+
+final Map<String, _BlockedPubkeysCacheEntry> _blockedPubkeysCaches = {};
+
+@visibleForTesting
+void resetBlockedPubkeysCacheForTests() => _blockedPubkeysCaches.clear();
+
 BlockedPubkeysState useBlockedPubkeys(String accountPubkey, {int refreshKey = 0}) {
-  final blockedPubkeys = useState<Set<String>>({});
-  final isLoading = useState(true);
+  final cache = _blockedPubkeysCaches.putIfAbsent(
+    accountPubkey,
+    _BlockedPubkeysCacheEntry.new,
+  );
+  final blockedPubkeys = useState<Set<String>>(cache.blockedPubkeys);
+  final isLoading = useState(!cache.hasLoaded);
   final error = useState<String?>(null);
   final manualRefreshKey = useState(0);
 
   useEffect(() {
     var cancelled = false;
-    isLoading.value = true;
+    if (!cache.hasLoaded) isLoading.value = true;
 
     Future<void> fetchBlockedPubkeys() async {
       try {
         final entries = await mute_list_api.getBlockedUsers(accountPubkey: accountPubkey);
         if (cancelled) return;
-        blockedPubkeys.value = entries.map((entry) => entry.mutedPubkey).toSet();
+        final next = entries.map((entry) => entry.mutedPubkey).toSet();
+        cache.blockedPubkeys = next;
+        cache.hasLoaded = true;
+        blockedPubkeys.value = next;
         error.value = null;
       } catch (e, st) {
         if (cancelled) return;
--- a/lib/hooks/use_chat_list.dart
+++ b/lib/hooks/use_chat_list.dart
@@ -139,6 +139,12 @@ ChatListResult _useChatList(
     [pubkey, refreshKey.value, archived],
   );
 
+  // useStream drives rebuilds via subscription side-effect; state lives in
+  // `cache` so it survives screen remounts. `isLoading` is gated on whether
+  // we already have a snapshot in cache, not on stream connection state —
+  // this is what lets the list show stale chats without a spinner while a
+  // fresh subscription is warming up (e.g. after navigating back, or after
+  // a manual refresh).
   useStream(stream, initialData: cache.chatMap);
   final isLoading = !cache.hasInitialSnapshot || blockedState.isLoading;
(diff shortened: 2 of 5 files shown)


lib/hooks/use_chat_list.dart
erskingardner, line 164: i think this is the intent of the PR - which is to show data when you go back and lazily make updates. BUT, I wanted to double check this. 

> When the user calls `refresh()`, `refreshKey` increments which re-subscribes to the stream, but `cache.hasInitialSnapshot` is never reset to `false`. This means `isLoading` stays `false` during the entire gap between the old subscription closing and the new `initialSnapshot` arriving. The visible effect is that after a manual refresh the list will briefly show the stale (pre-refresh) chats as if they were current, with no loading indicator to signal that data is being refreshed. Before this PR the old `ConnectionState.waiting` guard would naturally show the spinner during that window.
> 
> This may be acceptable if the intent is always "show cached, update lazily," but it would be worth an explicit decision and — if "stale is fine" — there should be a test that verifies the stale chats are shown (not hidden) during a refresh, so the contract is locked in.

test/hooks/use_chat_list_test.dart
erskingardner: There are two of these in the file - another on line 488. It would be worth cleaning that up.

lib/hooks/use_chat_list.dart
```diff
@@ -77,56 +94,56 @@ ChatListResult _useChatList(
...
               },
             );
           });
     },
     [pubkey, refreshKey.value, archived],
   );
 
-  final snapshot = useStream(stream, initialData: <String, ChatSummary>{});
-  final isLoading = snapshot.connectionState == ConnectionState.waiting || blockedState.isLoading;
+  useStream(stream, initialData: cache.chatMap);
```
erskingardner, line 148: > The `useStream` return value is intentionally dropped — the hook drives rebuilds through the `useStream` side-effect, then reads from `cache.chatMap` directly. This is a valid pattern but it is non-obvious: a future reader will wonder why the snapshot is ignored. A short comment explaining the intent (`// useStream drives rebuilds; state lives in cache`) would be consistent with "really complex or hard to understand" exceptions in the style guide.

lib/hooks/use_blocked_pubkeys.dart
```diff
@@ -12,21 +13,38 @@ typedef BlockedPubkeysState = ({
...
+  );
+  final blockedPubkeys = useState<Set<String>>(cache.blockedPubkeys);
+  final isLoading = useState(!cache.hasLoaded);
   final error = useState<String?>(null);
   final manualRefreshKey = useState(0);
 
   useEffect(() {
     var cancelled = false;
-    isLoading.value = true;
+    if (!cache.hasLoaded) isLoading.value = true;
```
erskingardner, line 43: Again - I think this is intentional for the PR but it's very non-obvious looking at the code (in 6 months time) why this is set to do this. 

> After the initial load, `cache.hasLoaded` is `true`, so calling `refresh()` (which increments `manualRefreshKey`) never sets `isLoading` back to `true` during the async fetch. The caller (`_useChatList`) reads `blockedState.isLoading` to decide whether to suppress the chat list. This means after a block/unblock-triggered refresh, blocked content could be briefly visible while the new blocked-pubkeys list is still fetching, if the stream update and the async fetch race. This was the original behavior before caching, but the guard against it (`isLoading = true` at the top of the effect) has been weakened. Intentional or not, it should be explicit.

lib/hooks/use_chat_list.dart
dannym-arx: yep, absolutely my intention

ghost:
Ready to review this PR? Stage has broken it down into 4 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Cache blocked pubkeys at module scope](https://stagereview.app/marmot-protocol/whitenoise/pull/641/chapters/1) |
| 2 | [Cache chat list state at module scope](https://stagereview.app/marmot-protocol/whitenoise/pull/641/chapters/2) |
| 3 | [Expose cache reset for testing](https://stagereview.app/marmot-protocol/whitenoise/pull/641/chapters/3) |
| 4 | [Verify chat list persistence in tests](https://stagereview.app/marmot-protocol/whitenoise/pull/641/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/641">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit e8dcbf7780070be89664dd169d099e1703a2d654 on May 11, 2026 7:49am UTC.</sub>

pepina-dev (changes requested):
I don't like this approach. This PR is adding a shared app global state keyed by account pub key. This state lives out of the widget tree, so it's not widget state and then does not belong to a hook, it is a hidden global cache.

The pattern we had defined  for shared app state is providers, and we moved away from heavy provider caching when starting this repo. This introduces the same hidden global state complexity that we had in the old repo through a different door. 

My main concern is that we are adding a cache layer of state that is invisible from the app architecture and call site, that doesn't have the advantages that does has riverpod providers lifecycle and that will make harder to then replace with a fix from rust side.

I would like to first, handle the visibility of chat in chat lists and of messages in rust side and notify to the flutter side via streams. That way, the blocked pubkeys cache would become completely unnecessary. 

I have the impression that @jgmontoya has already a rough idea on how to fix this on rust side once the refactor branch is merged (cause those changes would benefit from having separate db per account, so that visibility in aggregated messages can be stored in the same messages table, for groups visibility in chat list something similar would apply). 

Having said that, we will probably need to add some kind of cache anyways to improve first renders, but I want to delay that as much as possible and think about it carefully. I think we can still move more complex logic to rust and see if after doing that the loading in first render keeps being an issue. Maybe after those improvements it's not something noticeable and we can actually use rust as the source of truth on screens without cache.

pepina-dev:
I'm not saying that we will be able to get rid of this spinner/loading time without cache, I don't think we really can. BUT my hope is that we could make that loading time fast so that it doesn't matter that on first frame it's not rendered cause it wouldn't be noticeable.

Maybe I'm being too optimistic,  but I think we should try to move as much as we can to the rust side and if it is still slow🙏🏻 , then ok, I surrender and I'll agree on adding this chatlist cache (in a provider).  But blocked pubkeys cache it's  something we should't need once we move messages and groups visibility to be stored in rust db.

dannym-arx:
sure no worries! I can close the PR for now if you prefer

pepina-dev:
> sure no worries! I can close the PR for now if you prefer

Thanks for understanding Danny! 🫶

PR #646, opened 2026-05-09
Fix WnSystemNotice dismiss after unmount

### LINKED ISSUE #645: WnSystemNotice: 'Null check operator used on a null value' from auto-hide Timer firing after dispose
## Symptom

A user's logcat shows an unhandled Dart exception during a flow that surfaces a `WnSystemNotice` toast and then quickly transitions away from the screen:

\`\`\`
05-08 15:55:19.420  E flutter : [ERROR:flutter/runtime/dart_vm_initializer.cc(40)]
                                Unhandled Exception: Null check operator used on a null value
05-08 15:55:19.420  E flutter : #0  AnimationController.stop
                                    (package:flutter/src/animation/animation_controller.dart:899)
05-08 15:55:19.420  E flutter : #1  AnimationController._animateToInternal
                                    (package:flutter/src/animation/animation_controller.dart:673)
05-08 15:55:19.420  E flutter : #2  AnimationController.reverse
                                    (package:flutter/src/animation/animation_controller.dart:521)
05-08 15:55:19.420  E flutter : #3  WnSystemNotice.build.handleDismiss
                                    (package:whitenoise/widgets/wn_system_notice.dart:111)
05-08 15:55:19.420  E flutter : #4  Timer._createTimer.<anonymous closure>
                                    (dart:async-patch/timer_patch.dart:18)
\`\`\`

It happened during a rapid sequence of three failed \"Add to group\" attempts (~3 s apart) where each surfaced an error toast and the user navigated between screens. The auto-hide Timer for a previous toast fired against a controller whose ticker had already been nulled out — `AnimationController.stop` reads `_ticker!` which throws when the controller has been disposed.

## Suspected cause

`lib/widgets/wn_system_notice.dart`:

\`\`\`dart
105    final autoHideTimer = useRef<Timer?>(null);
106
107    void handleDismiss() {
108      if (isDismissing.value) return;
109      isDismissing.value = true;
110      autoHideTimer.value?.cancel();
111      slideController.reverse().then((_) {        // <-- crash site
112        onDismiss?.call();
113      });
114    }

125    useEffect(() {
126      final duration = autoHideDuration ?? (_isTemporary ? _defaultAutoHideDuration : null);
127      if (duration != null && onDismiss != null) {
128        autoHideTimer.value = Timer(duration, handleDismiss);
129      }
130      return () => autoHideTimer.value?.cancel();
131    }, [autoHideDuration, _isTemporary, onDismiss]);
\`\`\`

The cleanup on line 130 cancels the Timer on dispose / dep change. But there is at least one window where the Timer has already fired and `handleDismiss` is queued (or running) before the cleanup runs — at that point `slideController.reverse()` executes against a disposed controller and `AnimationController.stop`'s `_ticker!` blows up.

There may also be a longer-running variant: `handleDismiss` fires *before* dispose, kicks off `slideController.reverse()`, and the widget is dismounted while the reverse animation is still ticking — same crash, different phase of the same bug.

## Suggested fix directions

1. **Guard `handleDismiss` against a disposed controller.** `AnimationController` has no public `isDisposed`, but we can capture a `mounted`-style flag with a `useRef<bool>` set to `false` in a cleanup, and bail out of `handleDismiss` if not mounted. (Or use the `useIsMounted` hook from `flutter_hooks` directly — that's exactly what it's for.)
2. **Don't capture a stale `handleDismiss`.** The Timer at line 128 captures the closure from the first build. A small refactor — extract `handleDismiss` into a `useCallback` and have the Timer invoke through a stable indirection — keeps the cancel/dismiss pair in sync across rebuilds.
3. **Wrap the `.then(...)` chain.** Even if everything else is correct, `slideController.reverse().then(...)` is exposing dispose-during-animation. Use `vsync`-aware navigation guards or check status before calling `onDismiss`.

The minimum fix is probably (1) — wrap the body of `handleDismiss` in an `if (!isMounted()) return;` check.

## Repro

Hard to reproduce deterministically; the captured trace happened during this sequence:

1. Trigger an action that surfaces a `WnSystemNotice` error toast (e.g., the failing add-to-group flow that produced the surrounding log entries).
2. Within ~3 s, navigate away from the screen so the parent widget is dismounted while the toast's auto-hide Timer is still pending or its dismiss animation is mid-flight.
3. The Timer fires into a disposed controller — unhandled exception.

A targeted test: mount a `WnSystemNotice` with `autoHideDuration: Duration(milliseconds: 50)`, pump 30 ms, dispose the host widget, then `pump(100)` — should reproduce.

## Severity

Doesn't crash the app but produces an unhandled exception in the Dart VM and — depending on how Flutter's error zone is configured — may or may not bubble to FlutterError. Easy fix, worth doing.

## Source

Caught in a real user's logcat during an unrelated investigation. Same log surfaced the `whitenoise-rs` `add_members` issue: marmot-protocol/whitenoise-rs#813.

erskingardner:
## Summary
- Guard WnSystemNotice dismiss callbacks with context.mounted before touching hook state or completing the reverse animation
- Add a regression test for a queued dismiss callback firing after the notice has unmounted

Fixes #645

## Tests
- flutter test test/widgets/wn_system_notice_test.dart --plain-name 'ignores queued dismiss callback after unmount'
- flutter test test/widgets/wn_system_notice_test.dart
- just test-flutter-quiet
- just precommit

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/646">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented a crash where dismiss or auto-hide actions could run after a system notice was removed, ensuring dismiss callbacks only run while the notice is mounted.

* **Tests**
  * Added tests confirming dismissal and auto-hide timers are safely ignored if the notice is unmounted, preventing exceptions and unintended side effects.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/646)

### DIFF
--- a/lib/widgets/wn_system_notice.dart
+++ b/lib/widgets/wn_system_notice.dart
@@ -105,10 +105,12 @@ class WnSystemNotice extends HookWidget {
     final autoHideTimer = useRef<Timer?>(null);
 
     void handleDismiss() {
+      if (!context.mounted) return;
       if (isDismissing.value) return;
       isDismissing.value = true;
       autoHideTimer.value?.cancel();
       slideController.reverse().then((_) {
+        if (!context.mounted) return;
         onDismiss?.call();
       });
     }
--- a/test/widgets/wn_system_notice_test.dart
+++ b/test/widgets/wn_system_notice_test.dart
@@ -473,6 +473,54 @@ void main() {
 
       expect(dismissCount, 1);
     });
+
+    testWidgets('ignores queued dismiss callback after unmount', (tester) async {
+      var dismissed = false;
+      await mountWidget(
+        WnSystemNotice(
+          title: 'Auto-hide while navigating',
+          variant: WnSystemNoticeVariant.dismissible,
+          onDismiss: () => dismissed = true,
+        ),
+        tester,
+      );
+
+      await tester.pumpAndSettle();
+      final dismissGesture = tester.widget<GestureDetector>(
+        find.ancestor(
+          of: find.byKey(const ValueKey('systemNotice_actionIcon')),
+          matching: find.byType(GestureDetector),
+        ),
+      );
+      final queuedDismiss = dismissGesture.onTap;
+      expect(queuedDismiss, isNotNull);
+
+      await tester.pumpWidget(const SizedBox.shrink());
+      queuedDismiss!();
+      await tester.pump();
+
+      expect(tester.takeException(), isNull);
+      expect(dismissed, isFalse);
+    });
+
+    testWidgets('cancels pending auto-hide timer on unmount', (tester) async {
+      var dismissed = false;
+      await mountWidget(
+        WnSystemNotice(
+          title: 'Auto-hide timer while navigating',
+          autoHideDuration: const Duration(milliseconds: 50),
+          onDismiss: () => dismissed = true,
+        ),
+        tester,
+      );
+
+      await tester.pump(const Duration(milliseconds: 30));
+      await tester.pumpWidget(const SizedBox.shrink());
+      await tester.pump(const Duration(milliseconds: 100));
+
+      expect(tester.takeException(), isNull);
+      expect(dismissed, isFalse);
+    });
   });
 
   group('Expand/Collapse Animation', () {


ghost:
Ready to review this PR? Stage has broken it down into 2 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Guard dismiss callbacks with mount checks](https://stagereview.app/marmot-protocol/whitenoise/pull/646/chapters/1) |
| 2 | [Verify dismissal behavior after unmount](https://stagereview.app/marmot-protocol/whitenoise/pull/646/chapters/2) |

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/646">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 8fe8a08642aff273b494fbbde65e0a983db3cbd7 on May 11, 2026 6:36am UTC.</sub>

mubarakcoded (approved):

PR #647, opened 2026-05-09
Add Chinese localizations

stupidloud:
## Summary
- Add Chinese fallback localization zh required by Flutter for script-specific Chinese locales
- Add Simplified Chinese localization zh_Hans
- Add Traditional Chinese localization zh_Hant
- Regenerate Flutter localizations for the new Chinese locales

## Validation
- Ran ./scripts/validate-locales-keys.sh
- All locales validated successfully

## Notes
- zh_Hant was translated separately for Traditional Chinese users rather than mechanically converting from Simplified Chinese.
- This PR avoids unrelated generated formatting churn in existing locale files.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/647">
  <picture>
    Open in Stage
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added full Simplified and Traditional Chinese localizations and exposed distinct language options ("简体中文" and "繁體中文") in language settings.

* **Bug Fixes**
  * Improved locale handling so the app treats Simplified vs. Traditional Chinese as separate, consistent choices.

* **Tests**
  * Expanded tests to cover locale selection, display names, and notifier behavior for both Chinese variants.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/647)


dannym-arx:
Thank you so much for your contribution! 

No one on the WN team speaks chinese, so we're going to trust your translations. I checked some of them with kagi translate, they seem to be correct, but I have no way to know this isn't completely AI translated :slightly_smiling_face:

I will approve this PR, but can't merge until every commit is signed. and I think @pepina-dev or @cypherpinkdev should take a look

dannym-arx (dismissed):

pepina-dev (changes requested):
Thank you so much for your contribution!! 🎉 

In locale provider, the `_languageCodeToRust` method needs to have chinese added. Right now, both Chinese variants will get stored in rust crate on english and then language is reset to english on restart. I tried it and in fact it happens: I see the app in chinese, but once I close it and reopen it, is back to english

The main issue is that the language enum lives on the whitenoise rust crate 🦀, which doesn't include chinese yet. To support it we would need to add the two chinese variant there, merge it and bump it here. 

Something that worries me about this approach @dannym-arx on the rust crate is that this means every new language will require first a PR to the rust crate and then a PR here, which adds friction for new contributors that want to help with translations. Maybe we could handle the language code as a string instead of enum in the rust crate? Or is there a reason why we want this to be an enum?

dannym-arx:
@pepina-dev he did make a PR on the rust crate btw, he just didn't update the rev here!

I do agree with you tho, 💯 language should just be a string in rust, and we should keep an enum of allowed values in flutter, just defaulting to English if the value is invalid. I'll make a small PR for that in a bit

pepina-dev:
> @pepina-dev he did make a PR on the rust crate btw, he just didn't update the rev here!
> 
> I do agree with you tho, 💯 language should just be a string in rust, and we should keep an enum of allowed values in flutter, just defaulting to English if the value is invalid. I'll make a small PR for that in a bit

Oh sorry I hadn't realized that!

pepina-dev (approved):
Now it is persisted! 🎉   Thanks again @stupidloud !

PR #659, opened 2026-05-14
Render rich message content and Nostr mentions

erskingardner:
## Summary
- Add reusable rich message rendering for parsed message content, including tappable URLs and Nostr mention tokens.
- Render inline code and fenced code blocks with monospace styling, full-width block backgrounds, copy buttons, and copied feedback.
- Display known Nostr mentions as friendly `@name` text in chat bubbles and chat-list previews while preserving the underlying `nostr:npub...` content for sending/tapping.
- Add `@` mention suggestions in the composer for other chat members, with rich display text in the input and `nostr:npub...` stored/sent in the actual message.
- Keep mention-aware draft restoration and message sending wired through the chat input flow.

## Implementation Notes
- `MessageContentText` handles token-aware rendering for bubbles without taking ownership of chat-member lookup.
- Chat screens build the known mention display map from group members and pass it down to message bubbles.
- Chat-list previews do a small metadata-backed replacement pass because `ChatMessageSummary` currently contains raw content but not parsed content tokens.

https://github.com/user-attachments/assets/3f90630c-2847-4458-98b8-caa6e9a93999

CleanShot 2026-05-14 at 11 06 16

## Testing
- `flutter test test/widgets/chat_list_tile_test.dart`
- `just lint`
- `just test-flutter-quiet`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Rich message formatting: clickable links, inline & fenced code blocks with copy-to-clipboard and “Copied” feedback
  * Group chat @mentions: suggestions, friendly display names in messages/listings, and tap-to-open profile/navigation
  * Composer: mention-aware input that inserts/restores friendly mentions while preserving underlying identifiers for sending/drafts

* **Documentation**
  * Added localization strings for "Copy code" and "Copied" across languages

* **Tests**
  * Extensive widget tests for links, mentions, code blocks, copy UX, and draft persistence

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/659)


erskingardner:
@coderabbitai full review

lib/widgets/message_content_text.dart
```diff
@@ -0,0 +1,966 @@
...
+        });
+      } catch (_) {}
+    }
+
+    final copiedStyle =
+        Theme.of(context).textTheme.labelMedium?.copyWith(
+          color: Colors.white,
+          fontWeight: FontWeight.w700,
+        ) ??
+        const TextStyle(color: Colors.white, fontWeight: FontWeight.w700);
```
pepina-dev, line 256: prefer  app typography styles for sizing and weights and design system colors instead of using white directly.

lib/widgets/chat_message_bubble.dart
```diff
@@ -147,6 +171,7 @@ class ChatMessageBubble extends StatelessWidget {
       contentMaxLines: contentMaxLines,
       bubbleWidthFactor: bubbleWidthFactor,
       forceTightHeight: forceTightHeight,
+      onNostrTap: onNostrTap ?? (uri) => _openNostrProfile(context, uri),
```
pepina-dev, line 174: 🤖 My agent flagged this: 
> Gesture recognizer churn on every rebuild. This creates a new closure every build. It flows through to MessageContentText.useMemoized's dependency list — so _buildSpanState runs on every rebuild, creating and disposing
>  
> TapGestureRecognizer objects for every URL/mention in every visible message. With a real chat history of 50+ link-containing messages, this causes scroll jank.

lib/widgets/message_content_text.dart
```diff
@@ -0,0 +1,966 @@
...
+        _ContentRun(
+          text: content.substring(cursor, markerStart),
+          type: _ContentRunType.text,
+          start: cursor,
+          end: markerStart,
+        ),
+      );
+    }
+
+    if (_startsWithAt(content, markerStart, '```')) {
```
pepina-dev, line 520: I am a little bit confused on why some things are handled as tokens parsed in the rust side but others, like code blocks are parsed here. To me both seem like both are for display concerns and I don't understand why some leave in one side and one on the other.

lib/widgets/chat_list_tile.dart
```diff
@@ -25,6 +28,7 @@ import 'package:whitenoise/widgets/wn_chat_status.dart';
 import 'package:whitenoise/widgets/wn_icon.dart';
 
 final _logger = Logger('ChatListTile');
+final _nostrNpubUriPattern = RegExp(r'nostr:npub1[0-9a-z]+', caseSensitive: false);
```
pepina-dev, line 31: 🤖 My agent commented this:

> `_hasPreviewNostrUri` is case-sensitive, the regex is not
> 
> **File:** `lib/widgets/chat_list_tile.dart:61 and 95`
> 
> ```dart
> final _nostrNpubUriPattern = RegExp(r'nostr:npub1[0-9a-z]+', caseSensitive: false);
> // ...
> bool _hasPreviewNostrUri(String text) => text.contains('nostr:npub1');  // case-sensitive fast path

lib/widgets/chat_list_tile.dart
```diff
@@ -60,6 +64,44 @@ typedef _TileDisplay = ({
...
+      if (displayName == null || displayName.isEmpty) return;
+      displayNames[uri] = '@$displayName';
+    }),
+  );
+  return displayNames;
+}
+
+bool _hasPreviewNostrUri(String text) => text.contains('nostr:npub1');
+
+String _replacePreviewNostrMentions(String text, Map<String, String> displayNamesByUri) {
```
pepina-dev, line 97: The rust crate doesn't have this info in the already processed message tokens? like to know the tokens of  the last message in the rust side and if they are mention or not?

lib/widgets/message_content_text.dart
```diff
@@ -0,0 +1,966 @@
...
+        end: sliceEnd,
+        target: segment.target,
+      ),
+    );
+  }
+
+  return result;
+}
+
+int _min(int a, int b) => a < b ? a : b;
```
pepina-dev, line 676: [sug] we already user dart:math with min method in other places. I don't really care between this or the other, but it's more consistent to use the same one.

lib/widgets/chat_list_tile.dart
pepina-dev: I am not really sure it this is really an issue, but in case it is, this is a failing test:

```dart
testWidgets('resolves Nostr mentions with uppercase URIs in last message preview', (
        tester,
      ) async {
        _api.welcomerMetadata = const FlutterMetadata(displayName: 'Bob', custom: {});
        await pumpTile(
          tester,
          _chatSummary(lastMessageContent: 'This is for nostr:${testNpubB.toUpperCase()}'),
        );
        await tester.pumpAndSettle();

        final item = tester.widget<WnChatListItem>(find.byType(WnChatListItem));
        expect(item.subtitle, 'This is for @Bob');
      });
```

pepina-dev (changes requested):
This PR is too big and complex to review to feel confident enough that I'm not missing something 😰 

🤔 My main doubt/concern with this PR is why some things are parsed in rust crate (with messages tokens) and others are parsed here in flutter side. I think having all the parsing in message token in rust side would be more consistent, but maybe I am missing something. 

✅ Links work great across different devices

My agent raised an issue related to messages bubbles rebuild that I think it's worth checking 🤖  also,  here are some other issues I found:

- In chat screen, tags only look like tags in my messages, but I can't see tags from messages sent from other devices. 

Image

- When I try to tag someone in a message with image, I can't select the user cause it is above the image, I think that maybe it should appear below the images preview

Image

- UX: In whatsapp, when I tap a user tag ina group, I see the same info that  I would have seen when tapping that user from the members list. I mean not showing this start chat at the bottom on tap but the member info screen

|What now shows on tapping a tag| What I think it should show instead|
|----|----|
|Image| Image|

- In chat list code blocks look a bit weird: 

Image

dannym-arx:
@erskingardner I was gonna add markdown parsing (#406) (including nostr: links and bare npubs) first, implemented in rust, which then flutter would just take the output of and render. I think it would be better to do that first, then we do the @ mention dialog afterwards, once that has landed.

lib/widgets/message_content_text.dart
erskingardner: the reason for this is that the tokens in rust come from rust-nostr and it doesn't do full MD parsing (nor do we want full md parsing yet). I think for a first pass this is completely fine and doesn't require a full parsing engine in rust for something small. in the future if/when we want to expand it, we'll move things.

pepina-dev:
> @erskingardner I was gonna add markdown parsing (#406) (including nostr: links and bare npubs) first, implemented in rust, which then flutter would just take the output of and render. I think it would be better to do that first, then we do the @ mention dialog afterwards, once that has landed.

Nice, that would make the logic in flutter side simpler. Thanks @dannym-arx

dannym-arx:
@erskingardner @pepina-dev Please take a look at #665

PR #661, opened 2026-05-14
Add native deep links for users, chats, and settings

erskingardner:
## Summary
- Added custom URI deep-link support for `whitenoise://` and `whitenoise-staging://` without introducing any HTTP redirect/server dependency.
- Routed `user/<npub>`, `chat/<groupId>`, and settings deep links through the existing GoRouter/auth flow, including post-login continuation to protected targets.
- Preserved pending deep-link redirects through the relay-resolution login path as well as normal complete-login path.
- Registered production/staging schemes in Android and iOS app config so each installed flavor handles its own links; Android flavors now own their scheme explicitly without a default production fallback.
- Updated profile sharing so QR codes encode profile deep links while copy text still exposes the plain npub.
- Updated npub scanning to accept supported deep-link QR payloads, including chat links, as well as bare npubs.
- Added a `Copy Deep Link` affordance to the chat raw debug screen so QA can copy a real `chat/<groupId>` link from a live chat.
- Excluded `settings/sign-out` from supported settings deep links.

## Deep link contract
- `whitenoise://user/<npub>` / `whitenoise-staging://user/<npub>` opens the user profile/start-chat surface.
- `whitenoise://chat/<groupId>` / `whitenoise-staging://chat/<groupId>` opens a single chat.
- `whitenoise://settings/...` / `whitenoise-staging://settings/...` opens supported settings screens, excluding sign-out.

## Testing
- `flutter test test/utils/deep_links_test.dart`
- `flutter test test/providers/deep_link_provider_test.dart`
- `flutter test test/routes_test.dart`
- `flutter test test/screens/login_screen_test.dart`
- `flutter test test/screens/relay_resolution_screen_test.dart`
- `flutter test test/screens/share_profile_screen_test.dart`
- `flutter test test/screens/scan_npub_screen_test.dart`
- `flutter test test/screens/chat_raw_debug_screen_test.dart`
- `bash test/scripts/release_automation_config_test.sh`
- `plutil -lint ios/Runner/Info.plist ios/Runner/Runner.entitlements`
- `xmllint --noout android/app/src/main/AndroidManifest.xml`
- `just lint`
- `just test-flutter-quiet`
- `just precommit`

## Device validation
- Installed the staging build on Pixel 9a (`52281JEBF11805`).
- Verified Android resolves `whitenoise-staging://settings/notifications` to `org.parres.whitenoise.staging/org.parres.whitenoise.MainActivity`.
- Verified adb cold launch returns `Status: ok`.
- Manually verified settings, chat, and profile deep links on device.

## Summary by CodeRabbit

* **New Features**
  * Deep-link support to open the app to specific users, chats, or settings (staging/production schemes)
  * Shareable deep links and QR codes for profiles; scheme-aware “Copy Deep Link” in debug UI
  * Scanner recognizes user and chat deep links and navigates directly

* **Behavior Changes**
  * Protected deep-link targets route to login and return users to the intended destination (redirect preserved)
  * Post-auth flows now navigate to preserved deep-link destinations

* **Tests**
  * Expanded unit and widget tests covering parsing, routing, scanner, QR/copy, and platform wiring

* **Localization**
  * Added translations for the “Copy Deep Link” label across languages

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/661)


erskingardner:
@coderabbitai full review

lib/screens/share_profile_screen.dart
```diff
@@ -29,6 +31,10 @@ class ShareProfileScreen extends HookConsumerWidget {
     final pubkey = ref.watch(accountPubkeyProvider);
     final metadataSnapshot = useUserMetadata(context, pubkey);
     final npub = npubFromHex(pubkey);
+    final deepLinkScheme = ref.watch(deepLinkSchemeProvider).value ?? DeepLinks.productionScheme;
+    final profileLink = npub == null
+        ? null
+        : (npub: npub, uri: DeepLinks.userUri(npub, scheme: deepLinkScheme));
```
pepina-dev: 🐛   I tried scanning the QR from whitenoise from another device (not in this PR) and now it doesn't work,  so by adding this we are also breaking interoperability with previous whitenoise versions.

My agent flagged this too,  that this would be problematic for interoperability between clients.

pepina-dev: This is what my agent flagged 🤖 
> 
> Before this PR, the QR code encoded a bare `npub1...` string, which any Nostr client (Primal, Damus, Amethyst, etc.) can scan and recognize. After this PR, it encodes `whitenoise://user/npub1...`. A user who scans this with any app other than Whitenoise receives an unusable URL.
> 
> The scan logic in `scan_npub_screen.dart` falls through to the npub path if the deep link parse fails, so Whitenoise's own scanner still works — but this only helps if both sender and receiver use Whitenoise.
> 
> Notably, the **copy card still copies the bare npub**, creating an inconsistency: what you copy ≠ what the QR encodes.
> 
> **How to fix (choose one):**
> - **Option A (recommended):** Keep the QR encoding `npub1...` for maximum interoperability. Add a separate "Copy Deep Link" button below the QR for sharing within Whitenoise.
> - **Option B:** Encode `nostr:npub1...` (NIP-21 prefix), which many Nostr clients already handle, and also make `scan_npub_screen` recognise the `nostr:` prefix.
> 
> If the intent is deliberately Whitenoise-only QR, document it and add an in-screen label so users understand the QR is for Whitenoise only.

lib/screens/chat_raw_debug_screen.dart
```diff
@@ -187,6 +193,40 @@ class _DebugHeader extends StatelessWidget {
...
+                  mainAxisSize: MainAxisSize.min,
+                  children: [
+                    Icon(
+                      Icons.link_rounded,
+                      size: 14.w,
+                      color: colors.backgroundContentSecondary,
+                    ),
+                    SizedBox(width: 6.w),
+                    Text(
+                      'Copy Deep Link',
```
pepina-dev: There translation is missing

lib/screens/chat_raw_debug_screen.dart
```diff
@@ -187,6 +193,40 @@ class _DebugHeader extends StatelessWidget {
...
+                decoration: BoxDecoration(
+                  color: colors.fillSecondary,
+                  borderRadius: BorderRadius.circular(8.r),
+                  border: Border.all(color: colors.borderTertiary),
+                ),
+                child: Row(
+                  mainAxisSize: MainAxisSize.min,
+                  children: [
+                    Icon(
+                      Icons.link_rounded,
```
pepina-dev: use `WnIcons`

lib/screens/share_profile_screen.dart
```diff
@@ -29,6 +31,10 @@ class ShareProfileScreen extends HookConsumerWidget {
     final pubkey = ref.watch(accountPubkeyProvider);
     final metadataSnapshot = useUserMetadata(context, pubkey);
     final npub = npubFromHex(pubkey);
+    final deepLinkScheme = ref.watch(deepLinkSchemeProvider).value ?? DeepLinks.productionScheme;
+    final profileLink = npub == null
```
pepina-dev: I don't think this one is a blocker, but related to defaulting to production scheme, it left a suggestion 🤖 :

>  🟡 **SUGGESTION**  — `deepLinkSchemeProvider` silently falls back to production scheme during load
> 
> **File:** `lib/screens/share_profile_screen.dart:34`, `lib/screens/chat_raw_debug_screen.dart:172`
> 
> ```dart
> final deepLinkScheme = ref.watch(deepLinkSchemeProvider).value ?? DeepLinks.productionScheme;
> ```
> 
> `deepLinkSchemeProvider` is a `FutureProvider` that calls `PackageInfo.fromPlatform()`. While that future resolves, `.value` is null, and the fallback `productionScheme` is used. On a staging build, the QR code or debug link briefly shows a `whitenoise://` link instead of `whitenoise-staging://`.
> 
> In practice, `PackageInfo.fromPlatform()` resolves almost immediately, but it's still technically incorrect. Consider using `AsyncValue.when` to show a loading state or defer rendering until the scheme is known:
> ```dart
> final deepLinkSchemeAsync = ref.watch(deepLinkSchemeProvider);
> if (!deepLinkSchemeAsync.hasValue) return const SizedBox.shrink();
> final deepLinkScheme = deepLinkSchemeAsync.value!;
> ```

lib/screens/share_profile_screen.dart
pepina-dev: Maybe we could have the qr code with the npub and also allow to show the qr code with the deeplink, but I feel the previous QR is still needed for past versions and other clients

lib/screens/share_profile_screen.dart
```diff
@@ -7,8 +7,10 @@ import 'package:whitenoise/hooks/use_system_notice.dart';
 import 'package:whitenoise/hooks/use_user_metadata.dart';
 import 'package:whitenoise/l10n/l10n.dart';
 import 'package:whitenoise/providers/account_pubkey_provider.dart';
+import 'package:whitenoise/providers/deep_link_provider.dart';
 import 'package:whitenoise/routes.dart';
 import 'package:whitenoise/theme.dart';
+import 'package:whitenoise/utils/deep_links.dart';
```
pepina-dev, line 13: This is super weird cause I don't see any change in code that directly breaks it, but the style of QR codes is back to sharp borders, which was recently fixed in one o the ui polish PRs emir did

|master| this branch|
|----|----|
|Image|Image|

lib/routes.dart
```diff
@@ -698,4 +711,26 @@ abstract final class Routes {
...
+    if (redirect == null) return _login;
+    return Uri(path: _login, queryParameters: {'redirect': redirect}).toString();
+  }
+
+  static String _relayResolutionLocation({String? redirect}) {
+    if (redirect == null) return _relayResolution;
+    return Uri(path: _relayResolution, queryParameters: {'redirect': redirect}).toString();
+  }
+
+  static String? _safeRedirect(String? location) {
```
pepina-dev, line 725: My agent flagged that routes tests don't have tests for this checks. I think it makes sense to add some. I know this method is private and that's fine, but maybe it could be added on goToPostAuthDestination tests?

pepina-dev (commented):
My main concern is that npub QR scanning doesn't work after this changes for previous app versions, and won't work for users with other clients... so I think we should keep both qr codes and have a way to switch between them. 

I personally like how primal allows this very easily in a single screen to scan pubkey and ln address... maybe we could copy that pattern (if @vladimir-krstic aggrees)

Image

vladimir-krstic:
@pepina-dev  Yes, we will pretty much copy this.

erskingardner:
I don't think this is worth worrying about, given it doesn't break groups. People can and should update. We should be more pushy on updates while we're quickly adding new things.

lib/screens/share_profile_screen.dart
erskingardner: the rounded corners look broken to me. I think I'm going to stick with the way it is.

lib/routes.dart
erskingardner: Addressed in 85476532 by adding goToPostAuthDestination coverage through the login success path. The tests now keep the valid /chats/:groupId redirect case and cover fallback to the chat list for blank, absolute, authority-bearing, relative-path, and public-route redirects, which exercises the private _safeRedirect checks via the public post-auth flow.

PR #665, opened 2026-05-15
Add markdown rendering for chat messages

### LINKED ISSUE #406: Render Markdown inline and support .md file previews in groups
## Idea

Support Markdown rendering in two places:

1. **Inline messages** - Render Markdown formatting (bold, italic, code blocks, links, lists) directly in chat messages.
2. **Uploaded .md files** - When a `.md` file is shared in a group, render a preview of its contents instead of showing it as a raw file download.

This would make group conversations more readable and useful for sharing structured content, documentation, and notes.

dannym-arx:
marmot

https://github.com/user-attachments/assets/813ae5ee-daef-4bdf-aec6-cd0f390a2198

> Heads up: most of the diff is FRB codegen. Net total is ~7800 added lines; ~4300 of that is autogenerated `lib/src/rust/` and `rust/src/frb_generated.rs`, regenerated whenever the bridge surface changes. The hand-written review surface is ~1500 lines plus ~1700 lines of tests. Skim the codegen, read the rest.

This PR consumes the new `whitenoise_markdown::Document` AST from whitenoise-rs (pinned at `f421a8ad`). Chat bubbles now render CommonMark, GFM, and the nostr-extension format. Plain-text messages render byte-identically to before.

The marmots have learned to read bold text.

## What's in here

- `WnMarkdownText` walks the parsed AST and renders every Block and Inline variant.
- Plain-text fast path: single-paragraph messages with only Text/SoftBreak/HardBreak inlines skip the markdown renderer entirely. They keep the existing inline-timestamp layout.
- `@npub` mentions render as underlined `@DisplayName` (Vlad's design call). Tap opens a new in-app `UserProfileShade` instead of bouncing out to the OS. Non-npub Nostr URIs (note, nevent, nprofile, naddr, nrelay) still route externally because they point at events, not people.
- `SoftBreak` renders as a newline. The CommonMark default is a space, which loses every Enter press in the composer.
- URL scheme allow-list: `http`, `https`, `mailto`, `nostr`, `tel`. Anything else gets no tap recognizer at the renderer layer; the bubble's tap handler reapplies the same check before `launchUrl`. The Rust parser does not recognize HTML, so `<script>` and similar pass through as literal text.
- Surface boundaries: chat-list previews, reply quotes, notifications, and the composer keep their existing plain-text rendering.

## Bridge style

The bridge mirrors each AST type through FRB. The AST is FRB-friendly: owned `String`, `Vec<Enum>` recursion, `Clone`, `Eq`. Generated Dart sealed classes give exhaustive switch and IDE help. A hand-rolled Dart JSON parser would be a second markdown implementation and would drift from upstream.

## Tests + manual QA

- 4498 Flutter tests + 75 Rust tests pass. Coverage 99.03%. `just precommit` clean.
- Manual checklist run locally: every Block and Inline variant, surface boundaries, security cases (`javascript:`, `data:`, raw HTML stay inert), nested containers, CJK and RTL mixed runs, mention-tap into shade and back without scroll jump, long-press preview shows resolved names.

## Known limitations

1. Cross-inline highlight matches are not highlighted in search results.
2. Math renders as italic monospace source; no LaTeX engine is used.
3. Inline markdown images render as `[image: alt]` text links, not embedded.
4. Mentions of users outside the chat's metadata cache fall back to truncated npub until metadata loads. Follow-up could subscribe to mentioned pubkeys.

## Incidental upstream commits

The rev bump pulled in two non-markdown commits: `917ad14` (handle_contact_list lock contention) and `585d31f` (MDK update + push-token gossip). Both required adding `disappearing_message_secs: None` to two struct literals in `rust/src/api/groups.rs`.

## Summary by CodeRabbit

* **New Features**
  * Rich Markdown rendering for messages (formatted text, headings, code, lists, tables, math)
  * In-chat mention suggestions with resolved display names and insertion support
  * Tappable links and Nostr mentions with safe handling and in-app navigation (including opening profiles/chats)

* **UX Improvements**
  * Quoted-message previews render Markdown when available
  * Start-chat can open as a non-opaque “shade” for smoother flows

* **Localization**
  * Added localized text for unsupported deep-link dialogs (multiple languages)

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/665?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lib/src/rust/api/markdown.dart
```diff
@@ -0,0 +1,216 @@
...
+  @override
+  int get hashCode => blocks.hashCode;
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is MarkdownDocument &&
+          runtimeType == other.runtimeType &&
+          blocks == other.blocks;
+}
```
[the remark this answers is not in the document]
dannym-arx: this is autogenerated code

dannym-arx:
After talking with @erskingardner, I'm gonna fold #659 and #671 into this PR

so moving this one back to draft!

dannym-arx:
@claude full review, be in depth
@coderabbitai full review, be in depth

lib/widgets/markdown_text.dart
```diff
@@ -0,0 +1,584 @@
...
+  /// Style for an npub mention: bold, in the user's per-pubkey color.
+  /// Same palette as the bubble's sender-name color. When the hex pubkey
+  /// can't be derived, falls back to the surrounding text color so the
+  /// mention is still bold but doesn't pretend to identify someone.
+  TextStyle _mentionStyle(TextStyle style, String? hex) {
+    final color = hex != null
+        ? AvatarColor.fromPubkey(hex).toColorSet(context.colors).content
+        : style.color;
+    return style.copyWith(color: color, fontWeight: FontWeight.w700);
+  }
```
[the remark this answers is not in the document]
dannym-arx: no it renders them how vlad said they should be: 
image

dannym-arx:
@coderabbitai do a full review, I need you to find the golden carrot in this PR, the golden carrot is the single biggest issue in the PR that should get fixed. every PR has a golden carrot

lib/screens/chat_screen.dart
```diff
@@ -164,6 +262,19 @@ class ChatScreen extends HookConsumerWidget {
...
+        final hex = hexFromNpub(npub);
+        if (hex == null) return null;
+        return resolveMentionDisplayNameRef.value(hex);
+      },
+      const [],
+    );
+    useEffect(() {
+      input.controller.displayNameForNpub = displayNameResolver;
+      return null;
+    }, [input.controller, displayNameResolver]);
```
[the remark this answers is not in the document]
dannym-arx: I have not seen that happen in my tests

lib/widgets/markdown_text.dart
```diff
@@ -0,0 +1,641 @@
...
+
+const Set<String> _safeUrlSchemes = {
+  'http',
+  'https',
+  'mailto',
+  'nostr',
+  'tel',
+  'whitenoise',
+  'whitenoise-staging',
+};
```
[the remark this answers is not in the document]
dannym-arx: no this is working as intented

widgetbook/pubspec.lock
```diff
@@ -447,10 +447,10 @@ packages:
     dependency: transitive
     description:
       name: flutter_rust_bridge
-      sha256: e87d6b9ee934dcd24a128ccb2bd91905d2d5fe5c06245d6a8f5477d4907a437a
+      sha256: "37ef40bc6f863652e865f0b2563ea07f0d3c58d8efad803cc01933a4b2ee067e"
       url: "https://pub.dev"
     source: hosted
-    version: "2.12.0"
+    version: "2.11.1"
```
pepina-dev, line 453: You need to upgrade flutter_rust_bridge_codegen to 2.12.0" first and then run just deps and just regenerate  so that here it doesn't get downgraded

lib/utils/mention_text_editing_controller.dart
```diff
@@ -0,0 +1,539 @@
...
+  _TrackedMention mention,
+  List<_TrackedMention> replacements,
+  String oldText,
+  String newText,
+) {
+  var delta = 0;
+  for (final r in replacements) {
+    if (r.end <= mention.start) {
+      // replacement is fully before; accumulate delta
+      final oldRangeLen = r.uri.length;
```
pepina-dev, line 413: My agent raised this 🤖 

> `_shiftPastReplacements` uses the wrong "old length"** (`mention_text_editing_controller.dart:403-430`).
> The helper computes `oldRangeLen = r.uri.length`, but `r` is a `_TrackedMention` whose `uri` is the *bare npub* form (`@npub1…`, ~63 chars). When called from `_replaceBareNpubs` this is correct *by coincidence*; when called from `_replaceKnownUris` (`:327`), `r.uri` is the picker URI (e.g. `@npub1alice...`) but the **old** text contained `@DisplayName` (display text from a previous resolver pass). With short display names this currently produces a correct delta because both writers happen to use `uri.length` as old, but if a future caller registers a target where `displayText.length > uri.length`, surviving mentions are silently dropped.
> *How to address:* change `_shiftPastReplacements` to take `(oldStart, oldEnd, newStart, newEnd)` tuples instead of reconstructing from `_TrackedMention`. Add a test where `displayText.length > uri.length`.

lib/widgets/markdown_text.dart
```diff
@@ -0,0 +1,651 @@
...
+  /// Return `null` if unknown — renderer falls back to truncated npub.
+  final String? Function(String hexPubkey)? mentionDisplayName;
+  final List<String> highlightQueries;
+  final Color? highlightColor;
+  final int? maxLines;
+  final bool softWrap;
+
+  @override
+  Widget build(BuildContext context) {
+    final activeRecognizers = useRef<List<TapGestureRecognizer>>([]);
```
pepina-dev, line 122: My agent raised this concern here 🤖 
> **`MarkdownText` `TapGestureRecognizer`s can leak past widget unmount** (`markdown_text.dart:122-141`).
> Stale recognizers from the previous build are disposed via `WidgetsBinding.instance.addPostFrameCallback`. If the widget is removed before the next frame (Navigator.pop during scroll, hot reload, fast list rebuild), the callback still runs but the recognizers held callbacks that captured `BuildContext` and closures referencing now-disposed objects.
> *How to address:* dispose stale recognizers **synchronously** at the start of `build` before reassigning `activeRecognizers.value`. Add a leak test that pumps the widget, unmounts, and asserts no listener leak (`leak_tracker`).

lib/widgets/markdown_text.dart
```diff
@@ -0,0 +1,651 @@
...
+List<MarkdownInline>? _firstInlinesInBlocks(List<MarkdownBlock> blocks) {
+  for (final block in blocks) {
+    final inlines = switch (block) {
+      MarkdownBlock_Paragraph(:final inlines) => inlines,
+      MarkdownBlock_Heading(:final inlines) => inlines,
+      MarkdownBlock_BlockQuote(:final blocks) => _firstInlinesInBlocks(blocks),
+      MarkdownBlock_List(:final items) =>
+        items.isEmpty ? null : _firstInlinesInBlocks(items.first.blocks),
+      MarkdownBlock_Table(:final header) => _joinCellInlines(header),
+      MarkdownBlock_CodeBlock(:final content) => [
```
pepina-dev, line 63: 🤖 Agent concern
> **Image alt-text and code-block previews are unbounded** (`markdown_text.dart:478-484`, `:63-68`).
> A peer can send ` alt=<10 KB>`, which renders as `[image: <10 KB>]` in the bubble; `firstParagraphInlines` will also lift a code-block's first line into the reply preview verbatim. Both are abusable for visual DoS (an oversized preview row breaking the chat list).
> *How to address:* truncate alt text to ~100 chars before rendering (`markdown_text.dart:478`). In `_firstNonEmptyLine` (`:85`), cap to ~120 chars and add an ellipsis. Add a test with 10 KB alt text.

lib/widgets/markdown_text.dart
```diff
@@ -0,0 +1,651 @@
...
+    }
+
+    final renderer = _MarkdownRenderer(
+      context: context,
+      baseStyle: baseStyle,
+      onLinkTap: onLinkTap,
+      onNostrTap: onNostrTap,
+      mentionDisplayName: mentionDisplayName,
+      highlightQueries: highlightQueries,
+      highlightColor: highlightColor ?? Colors.yellow,
```
pepina-dev: Use colors set in theme

lib/widgets/chat_message_bubble.dart
```diff
@@ -83,6 +92,47 @@ class ChatMessageBubble extends StatelessWidget {
...
+    if (scheme == DeepLinks.productionScheme || scheme == DeepLinks.stagingScheme) {
+      final target = DeepLinks.parse(uri);
+      if (target == null) {
+        if (context.mounted) await _showUnsupportedDeepLinkDialog(context);
+        return;
+      }
+      if (target.type == DeepLinkTargetType.user) {
+        final hex = hexFromNpub(uri.pathSegments.first);
+        if (hex != null && context.mounted) {
+          await StartChatScreen.show(context, userPubkey: hex);
```
pepina-dev, line 109: I had mentioned it in https://github.com/marmot-protocol/whitenoise/pull/659#pullrequestreview-4291375203 related to showing start chat screen on mentions... I think we should be consistent with that we show when tapping on member lists avatars. 

In whatsapp, when I tap a user tag in group, I see the same info that  I would have seen when tapping that user from the members list. I mean not showing this start chat at the bottom on tap but the member info screen

lib/widgets/markdown_text.dart
```diff
@@ -0,0 +1,651 @@
...
+  }
+
+  /// Renders the display name for an npub mention.
+  /// Looks up via [mentionDisplayName]; if unresolved, falls back to a
+  /// truncated bech32 ('npub1abcdefgh…wxyz'). No '@' prefix per design.
+  String _resolveMentionLabel(String bech32, String? hex) {
+    final lookup = mentionDisplayName;
+    if (lookup != null && hex != null) {
+      final name = lookup(hex);
+      if (name != null && name.isNotEmpty) {
```
pepina-dev, line 572: > 🤖 Display-name resolver doesn't trim whitespace** (`markdown_text.dart:577`). `name.isNotEmpty` passes through a single space. Compare with the controller (`:191`) which already trims — be consistent.
> *How to address:* change to `name.trim().isNotEmpty` and use the trimmed value.

lib/utils/mention_query.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  if (start >= caret || value.text[start] != '@') return null;
+  return ActiveMentionQuery(
+    start: start,
+    end: caret,
+    query: value.text.substring(start + 1, caret),
+  );
+}
+
+bool _isMentionBoundary(int codeUnit) =>
```
pepina-dev, line 35: > 🤖 `_isMentionBoundary` only accepts whitespace, so `@npub1...!` will refuse to auto-truncate. The Rust parser is more permissive.
> *How to address:* extend `_isMentionBoundary` to also accept `.`, `,`, `;`, `:`, `!`, `?`, `)`, `]`. Add tests.

lib/widgets/markdown_text.dart
pepina-dev: I'm not sure this is a real issue, it's hard to test

widgetbook/pubspec.lock
dannym-arx: I have no idea why this happened, thanks!

lib/widgets/markdown_text.dart
dannym-arx: I think the agent meant `10kb`, which I guess is real... but they can also send 10kb of text, there's no difference haha

image

image

lib/utils/mention_text_editing_controller.dart
dannym-arx: I'm not sure I understand this issue. 

image
image
image

lib/widgets/chat_message_bubble.dart
dannym-arx: I don't know if I agree... sorry! 

Because I should be able to mention anyone, not just people in the chat using their npub, and if we do that, we can't show the same screen as when we tap the members list. 

e.g. in a chat you're not in I can mention you

image

and tapping your name should show a way to message you directly, or add you to a group.

image

---

perhaps a compromise would be that if they're in the group we should the group member page in slate, and if they're not, we show the start chat screen?

erskingardner: I think that your compromise is probably reasonable. it's a little inconsistent but for good reason.

lib/widgets/markdown_text.dart
pepina-dev: lol haha ok so this one can be ignored. (Sorry bout this, I tried to filter irrelevant comments cause my agent left lots of  warnings as this PR was big)

lib/widgets/chat_message_bubble.dart
pepina-dev: I hadn't realized about the case of mentions of people external to the group. Then yes, for that case the start chat makes sense.

 I like the idea of showing the group member page if it is a member and the start chat if it is not, but at the same time we don't want to load members in every chat message bubble, that would make the chat rendering slow. It would probably require to handle it a more sophisticated way to avoid compromising performance.

TLDR: let's leave the start chat screen for all cases for now  and handle the case of group members in a separate issue

pepina-dev: https://github.com/marmot-protocol/whitenoise/issues/681

lib/utils/mention_text_editing_controller.dart
pepina-dev: Sorry about this, I've been trying to reproduce too now and I can't see any issue here 🙈. I tried to filter irrelevant comments cause my agent left lots of warnings , but this one should had been filtered

pepina-dev (approved):
LGTM ✅

PR #673, opened 2026-05-18
Implement iOS NSE notifications and Android push token plumbing

erskingardner:
## Summary
- Add iOS notification service extension support to replace blank APNS payloads with localized message previews.
- Wire APNS/FCM token collection through Flutter into `whitenoise-rs` for MIP-05 push registration.
- Keep Android foreground notification handling in place while adding FCM token and wakeup plumbing for later mode selection.
- Harden iOS startup storage/keychain migration so upgraded installs keep working across App Group and keychain changes.
- Refresh generated Flutter Rust Bridge bindings and update docs/agent guidance for the new install and notification flows.

## Testing
- `just precommit` passed.
- Ran focused Rust and Flutter tests covering push registration, notification handling, startup migration, and iOS keychain behavior.
- Validated staged installs and local device behavior on iOS and Android during manual notification testing.

## Summary by CodeRabbit

* **New Features**
  * Cross-platform push support: iOS APNS + Android FCM provider token flows, registration sync, background wake, and in-app permission/request flows
  * iOS Notification Service Extension to decrypt/replace notification content
  * Hooks/providers to trigger chat/message list refreshes from notification events

* **Documentation**
  * Added iOS device testing and push setup guidance

* **Chores**
  * Build/config updates and multi-language notification strings

* **Tests**
  * Expanded tests for push, notification handling, migration, and refresh behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/673?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lib/main.dart
```diff
@@ -53,31 +59,187 @@ Future<void> main() async {
...
+  required Directory from,
+  required Directory to,
+  required Future<Directory> Function(Directory from, String newPath) renameDirectory,
+  required Future<void> Function(Directory from, Directory to) copyDirectory,
+  required Future<void> Function(Directory directory) deleteSourceDirectory,
+}) async {
+  if (!from.existsSync()) return;
+
+  if (to.existsSync()) {
+    if (await _directoryHasFiles(Directory('${to.path}/data'))) return;
```
mubarakcoded: This migration needs to distinguish real migrated app data from App Group data that the NSE may have created. If a push arrives before the updated app is opened, the extension can create/open `AppGroup/whitenoise/data`, then this `to/data has files` check returns early, and the existing Documents database is never moved. That could leave the app using a fresh App Group store while the user's real data remains in Documents.

erskingardner: good catch! fixed in the latest

rust/src/api/mod.rs
```diff
@@ -139,6 +139,8 @@ pub use zapstore::*;
 
 #[frb]
 pub async fn initialize_whitenoise(config: WhitenoiseConfig) -> Result<(), ApiError> {
+    crate::ios_keyring::install_ios_keyring_store_if_needed()
```
dannym-arx, line 251: from my agent:

**iOS keyring access group likely reset on first init.**

This call passes no access group, but Swift already called `installBackgroundKeyringStore` from `AppDelegate.didFinishLaunchingWithOptions` with the value from `WNKeychainAccessGroup`. Since `set_default_store` is a `set`, this second call wins and the access group is reset to `None` on the main app process. NSE is unaffected (doesn't call `initialize_whitenoise`).

Net effect: the access-group plumbing works for NSE but is silently disabled on Runner after first init. Suggest removing the auto-install from `initialize_whitenoise` on iOS, or having it accept/forward the access group from Swift.

ios/WhiteNoiseNotificationServiceExtension/NotificationService.swift
```diff
@@ -0,0 +1,318 @@
...
+    }
+    guard let container = appGroupContainerURL() else {
+      NSLog("White Noise NSE App Group container unavailable")
+      return .failure("app group")
+    }
+
+    let baseDir = container.appendingPathComponent("whitenoise", isDirectory: true)
+    let dataDir = baseDir.appendingPathComponent("data", isDirectory: true).path
+    let logsDir = baseDir.appendingPathComponent("logs", isDirectory: true).path
+    try? FileManager.default.createDirectory(atPath: dataDir, withIntermediateDirectories: true)
```
dannym-arx: from my agent:

**Fresh-install NSE race could wipe a real DB.**

This `createDirectory` provisions `data/` in the App-Group container with no `data_version` marker. On a fresh install where APNS wakes the NSE before the user opens the app, the marker-less directory will be deleted on first app launch by `_moveWhitenoiseDirectoryIfNeeded` ("no marker → wipe"). Consider having the NSE drop the marker itself when it provisions the dir, so the migration heuristic doesn't treat NSE-created data as discardable.

rust/src/api/mod.rs
jgmontoya, line 251: My agent flagged this:

> `ios/Runner/AppDelegate.swift:243` installs the iOS keyring store with `wn_install_ios_background_keyring_store_with_access_group(...)`, which is exactly what the app and notification service extension need in order to share the database key through the configured keychain access group. But `initialize_whitenoise` then calls back into `rust/src/ios_keyring.rs:26`, which installs the same store with `None` for the access group. `rust/src/ios_keyring.rs:61` calls `keyring_core::set_default_store(...)`, and that API replaces the existing default store rather than preserving it.
> 
> That means the main app can end up using a no-access-group keychain store after Rust initialization, while the NSE uses the access-group store. The App Group database path is shared, but the encryption key is not, so background notification collection can fail to decrypt/read the same database. Fix this by making `install_ios_keyring_store_if_needed` preserve an already configured default store, or by passing the access group into `initialize_whitenoise` and installing one consistent store from Rust.
>

lib/main.dart
```diff
@@ -53,31 +59,182 @@ Future<void> main() async {
...
+  final documentsDir = await getApplicationDocumentsDirectory();
+  final documentsBaseDir = Directory('${documentsDir.path}/whitenoise');
+  if (!(isIOS ?? Platform.isIOS)) return documentsBaseDir;
+
+  try {
+    final appGroupContainerPath = await appGroupChannel.invokeMethod<String>(
+      _getAppGroupContainerPathMethod,
+    );
+    if (appGroupContainerPath == null || appGroupContainerPath.isEmpty) {
+      return documentsBaseDir;
```
jgmontoya: From my agent:

> `resolveWhitenoiseBaseDirectory` returns the Documents path when the iOS App Group channel returns null or an empty path, and the catch block at `lib/main.dart:178` does the same for any channel or migration error. The NSE, however, always reads `container/whitenoise/data` from the App Group in `ios/WhiteNoiseNotificationServiceExtension/NotificationService.swift:163`.
> 
> If signing, entitlements, or the native channel are wrong, the main app will keep running against Documents while the NSE reads an empty or separate App Group database. That hides the configuration failure and produces a split-brain storage state. For production iOS builds this should fail fast with a clear fatal error, or be guarded behind an explicit debug/test-only fallback so notification support cannot silently ship broken.
>

lib/main.dart
```diff
@@ -53,31 +59,182 @@ Future<void> main() async {
...
+        error,
+        stackTrace,
+      );
+      await retryDelay(_initializeWhitenoiseRetryDelay);
+    }
+  }
+}
+
+bool _isDatabasePoolTimeout(Object error) {
+  final message = error.toString().toLowerCase();
```
jgmontoya: > The database-pool retry path is matching a human-readable SQLx error string. That is fragile: a wording change disables the retry, and an unrelated error containing the same phrase would be retried. The Rust boundary still has access to structured errors before converting them into the flattened Flutter `ApiError`; classify this in Rust instead, either with a typed `ApiError` variant or by doing the retry around `Whitenoise::ensure_initialized` while matching `sqlx::Error::PoolTimedOut` / the database wrapper.

android/app/src/main/kotlin/org/parres/whitenoise/WhitenoiseApplication.kt
```diff
@@ -18,7 +18,7 @@ class WhitenoiseApplication : Application(), FlutterForegroundTaskLifecycleListe
         // by RebootReceiver / MY_PACKAGE_REPLACED.
         try {
             Keyring.initializeNdkContext(applicationContext)
-        } catch (t: Throwable) {
+        } catch (t: Exception) {
```
jgmontoya: > This branch narrowed the previous `Throwable` catches to `Exception`, but `Keyring.initializeNdkContext(...)` crosses a JNI/native-library boundary. `android/app/src/main/kotlin/io/crates/keyring/Keyring.kt:8` runs `System.loadLibrary("rust_lib_whitenoise")`, and load/link failures such as `UnsatisfiedLinkError` are `Error`s, not `Exception`s. In a background process entry point, that would crash before this log path runs.
> 
> At this boundary the broader catch is justified. Restore `Throwable` for the keyring initialization path, with a short comment that this is intentionally limited to JNI/library loading. Consider the same for plugin registration if the foreground-task engine must remain fail-open.
>

lib/providers/notification_provider.dart
```diff
@@ -34,27 +63,37 @@ final notificationServiceProvider = Provider<NotificationService>((ref) {
 });
 
 final notificationListenerProvider = Provider.autoDispose<void>((ref) {
-  if (!Platform.isAndroid) return;
+  if (!notificationsSupported()) return;
```
jgmontoya, line 69: > The branch adds `routePendingTap`, iOS `consumePendingNotificationTap`, and the local-notification display gate under this existing coverage-ignore block. Those functions are testable and already have tests, but the coverage gate will not count or protect them while they remain inside the ignored region. Given the project rule against `coverage:ignore` for reachable code, narrow the ignore to only the platform/Riverpod wiring that cannot reasonably be covered, or move the pure helpers outside the ignored block.

I think we should remove the `//coverage:ignore-start` from line 50 and // coverage:ignore-end (line 279). This is adding a lot of code in between.

ios/WhiteNoiseNotificationServiceExtension/NotificationService.swift
```diff
@@ -0,0 +1,318 @@
...
+      return groupName ?? "Secure group invite"
+    }
+    if isDm {
+      return sender.displayName ?? "White Noise"
+    }
+    return groupName ?? "White Noise"
+  }
+
+  var body: String {
+    let senderName = sender.displayName ?? "Someone"
```
jgmontoya: > The extension adds localized fallback strings, but the decrypted notification title/body templates are hard-coded in English (`"Someone"`, `"Secure chat invite"`, `"invited you..."`). Since NSE-rendered notifications are user-visible, move these templates into `Localizable.strings` with format arguments so iOS remote notifications match the rest of the localized app.

lib/screens/chat_list_screen.dart
```diff
@@ -181,8 +182,13 @@ class ChatListScreen extends HookConsumerWidget {
     final isOffline = ref.watch(offlineProvider).value ?? false;
     final typography = context.typographyScaled;
     final pubkey = ref.watch(accountPubkeyProvider);
-    final chatListResult = useChatList(pubkey);
-    final archivedChatListResult = useChatList(pubkey, archived: true);
+    final chatListRefreshToken = ref.watch(chatListRefreshProvider);
```
pepina-dev: Why do we need this provider? You know I'm not a fan of adding providers for widget state... so I'll push back on this one

lib/hooks/use_chat_messages.dart
```diff
@@ -109,7 +110,7 @@ ChatMessagesResult useChatMessages(
     paginationVersion.value = 0;
     requestTokenRef.value++;
     return null;
-  }, [groupId, pubkey, hiddenPubkeysKey]);
+  }, [groupId, pubkey, hiddenPubkeysKey, refreshToken]);
```
pepina-dev: Why is this needed??

lib/providers/chat_list_refresh_provider.dart
```diff
@@ -0,0 +1,14 @@
...
+class ChatListRefreshNotifier extends Notifier<int> {
+  @override
+  int build() => 0;
+
+  void requestRefresh() {
+    state++;
+  }
+}
+
+final chatListRefreshProvider = NotifierProvider<ChatListRefreshNotifier, int>(
```
pepina-dev: Ok so my comment got stale, I don't understand why we need a provider to refresh the chat list screen. Providers are for global state, widget state lives in hooks

lib/screens/chat_screen.dart
```diff
@@ -155,6 +156,7 @@ class ChatScreen extends HookConsumerWidget {
     final typography = context.typographyScaled;
     final pubkey = ref.watch(accountPubkeyProvider);
     final isOffline = ref.watch(offlineProvider).value ?? false;
+    final chatListRefreshToken = ref.watch(chatListRefreshProvider);
```
pepina-dev: This is confusing: why a chat list refresh provider should affect a single chat screen? By it's name I thought it only affected the chatlist

lib/services/push_token_service.dart
```diff
@@ -0,0 +1,135 @@
...
+      _logger.warning(
+        'Failed to request notification permission',
+        error,
+        stackTrace,
+      );
+      return false;
+    }
+  }
+
+  @visibleForTesting
```
pepina-dev: I don't like visible for testing approach and I try to avoid it as much as possible. With this we end up testing the private methods instead of testing the public interface behavior.

pepina-dev (commented):
I don't understand why do we need to add a provider to refresh the chat list and chat screens... 🤔

pepina-dev (commented):
@erskingardner in this branch I can no longer run the app in release mode in my device... which I use all the time. Do you need to register my device ids again?

`flutter run --release -d 'mydevicename' --flavor staging`

Or do I need a secret? This is the error I get:
```

Could not build the precompiled application for the device.
Error (Xcode): No signing certificate "iOS Distribution" found: No "iOS Distribution" signing
certificate matching team ID "QL6TB358UB" with a private key was found.
```

I checked in apple developer and my device ID for my iphone 11 is registered in the team... So I don't know if maybe now you need to give me a new permission?

pepina-dev (commented):
My testing results on android:

- I don't get requested for notifications permissions... is that expected? I deleted the staging apk and reinstalled it but I still don't see anything. Tried this multiple times
- 🐛 Now on tapping (local? I guess if no permission was requested) notification for another account, accounts don't switch to the one that was supposed to receive the notification. As far as I remember, this was working in master

rust/src/api/mod.rs
erskingardner: Addressed: Rust no longer reinstalls the iOS keyring store during initialize; Swift/NSE own the access-group install.

erskingardner: Addressed: initialize no longer resets the Swift-installed keychain access group.

lib/main.dart
erskingardner: Addressed: missing/unavailable App Group lookup now fails fast instead of falling back to Documents.

ios/WhiteNoiseNotificationServiceExtension/NotificationService.swift
erskingardner: Addressed: decrypted NSE notification templates now use localized strings with format arguments.

lib/providers/notification_provider.dart
erskingardner: Addressed in 9074c7f1: narrowed the coverage ignores to platform/Riverpod/native routing glue, and kept the pure notification tap/channel helpers covered. `just coverage` is back at 99.03%.

lib/services/push_token_service.dart
erskingardner: Addressed in 9074c7f1: made the push-token method-call handler private again and updated tests to exercise native callbacks through the MethodChannel boundary.

ios/WhiteNoiseNotificationServiceExtension/NotificationService.swift
erskingardner: Addressed in 242f789e. The main-app migration now preserves App Group data when legacy Documents has no actual data files, and the NSE stamps the data_version marker after successfully collecting new notification data.

lib/main.dart
erskingardner: Addressed in 242f789e. The bridge now maps structured sqlx::Error::PoolTimedOut into a typed ApiError.databasePoolTimedOut, and Dart retries only that typed variant rather than matching error text.

PR #676, opened 2026-05-18
feat(blocked-users): complete block/unblock UX

mubarakcoded:
## Description

Completes the block/unblock feature with the two missing pieces:

1. **Hide the chat composer when the peer is blocked.** The "You blocked this user" banner already rendered in chats; only the input gate was missing.
2. **Settings → Privacy & Security → Blocked users.** A list of all blocked accounts, tap a row to open a profile-detail modal with an Unblock action. The detail opens over a blurred list following the existing `chat_info_screen` pattern (`opaque: false` + `WnOverlay(light)`).

Reuses existing hooks (`useBlockActions`, `useBlockedPubkeys`, `useUserMetadata`, `useRouteRefresh`) and shared widgets (`WnSlate`, `WnOverlay`, `WnChatInfoProfileCard`, `WnUserItem`). No Rust changes.

**Architectural note:** `BlockedUsersScreen`'s `WnSlate` uses a unique Hero tag (`'blocked-users-list-slate'`). `chat_info_screen` keeps the chat content visible behind by placing the message list *outside* its slate; the mock for this feature requires the list *inside* the slate, so the source-only tag asymmetry is the minimum-deviation way to keep the list visible behind the detail modal.

## Type of Change

- [ ] 🛠️ Bug fix
- [ ] ❌ Breaking change
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

## How I Tested This

- **Pixel 9 Pro XL** (Android, physical device, arm64-v8a): full walkthrough — block a peer in chat → composer hides → Settings → Privacy → View blocked users → tap a row → modal-over-blurred-list renders → tap Unblock → list refreshes via `useRouteRefresh` on pop.
- **Pixel_9_Pro AVD** (Android emulator, x86_64, Android 16): clean install + boot to login splash on a fresh AVD, same process as above.

Test suite: 4489/4489 pass. Coverage 99.40% global, 100% on the two new screens. `just precommit` green across all 9 gates.

## Screenshots

Screenshot_1779140812
Screenshot_1779140819
Screenshot_1779140847

UI changes 
- Settings → Privacy & Security with the new "Blocked users" section
- Blocked users list
- Blocked user profile detail (modal over blurred list)
- Chat with a blocked peer (composer hidden, banner visible)

## Checklist

- [x] `just precommit` passes
- [x] `CHANGELOG.md` updated
- [x] Screenshots added

## Summary by CodeRabbit

* **Documentation**
  * Updated changelog noting the blocked-users feature.

* **New Features**
  * Added Blocked Users management under Settings → Privacy & Security.
  * View blocked-users list, open blocked-user profiles, and unblock users.
  * Chat composer/input is hidden when a peer is blocked.

* **Localization**
  * Added translations for the blocked-users UI across multiple languages.

* **Tests**
  * Added widget tests covering blocked-users screens, unblock flows, and related UI states.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/676?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lib/screens/blocked_users_screen.dart
```diff
@@ -0,0 +1,122 @@
...
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final colors = context.colors;
+    final typography = context.typographyScaled;
+    final accountPubkey = ref.watch(accountPubkeyProvider);
+    final blocked = useBlockedPubkeys(accountPubkey);
+
+    useRouteRefresh(context, blocked.refresh);
+
+    final sortedPubkeys = blocked.blockedPubkeys.toList()..sort();
```
pepina-dev: This sorting by pubkey will look random for users. I think we should sort by the name we are showing to the user

lib/routes.dart
```diff
@@ -566,6 +587,16 @@ abstract final class Routes {
     GoRouter.of(context).push(_privacySecurity);
   }
 
+  static Future<void> pushToBlockedUsers(BuildContext context) {
```
pepina-dev, line 609: I don't see tests for this new methods

lib/screens/blocked_users_screen.dart
pepina-dev: And if the user has no name, then by npub in bech 32 format instead of hex, cause that's the one we are showing to the user

lib/screens/blocked_user_screen.dart
```diff
@@ -0,0 +1,170 @@
...
+                        onPublicKeyCopied: () => systemNotice.showSuccessNotice(
+                          context.l10n.publicKeyCopied,
+                        ),
+                        onPublicKeyCopyError: () => systemNotice.showErrorNotice(
+                          context.l10n.publicKeyCopyError,
+                        ),
+                      ),
+                      Gap(16.h),
+                      Container(
+                        key: const Key('blocked_user_detail_card'),
```
pepina-dev: This looks like a reimplementation of a collapsable WnSystemNotice. Please take a look to the chat screen where a similar notice is shown here. Also, in design I don't see that horizontal badding between this notice and the border of the slate

Image

lib/screens/blocked_user_screen.dart
```diff
@@ -0,0 +1,170 @@
...
+      userPubkey: userPubkey,
+    );
+    final systemNotice = useSystemNotice();
+    final isBannerCollapsed = useState(false);
+
+    Future<void> handleUnblock() async {
+      try {
+        await blockState.toggleBlock();
+        if (!context.mounted) return;
+        Routes.goBack(context);
```
pepina-dev: @vladimir-krstic I need your opinion here ... is going back in this case the expected behaviour for success on unblock? 

This  is not the usual navigation pattern in the app. We normally handle success or errors in the same screen by showing a wn system success notice. In chat screen, I don't see the success notice BUT after blocking/unblocking the button text changes and the app does not automatically navigates back.

pepina-dev (changes requested):
Please use design system component `WnSystemNotice` for this warning so that we don't end up with duplicated widgets that do the same.

lib/routes.dart
mubarakcoded: Nice catch! It has been added, thanks.

lib/screens/blocked_user_screen.dart
vladimir-krstic: You are right, we should not go back. Simply user profile card view when user is not blocked. 

So, from this screen: 
CleanShot 2026-05-19 at 22 15 08@2x

To this:
CleanShot 2026-05-19 at 22 16 47@2x

pepina-dev: Thanks @vladimir-krstic

lib/screens/blocked_users_screen.dart
```diff
@@ -29,10 +30,16 @@ class BlockedUsersScreen extends HookConsumerWidget {
 
     useRouteRefresh(context, blocked.refresh);
 
-    final sortedPubkeys = blocked.blockedPubkeys.toList()..sort();
+    final pubkeysKey = (blocked.blockedPubkeys.toList()..sort()).join(',');
```
pepina-dev: Why is this key needed? The double sort looks like unnecessary to me

lib/screens/blocked_users_screen.dart
```diff
@@ -0,0 +1,148 @@
...
+  }
+}
+
+Future<List<String>> _sortByDisplayName(Set<String> pubkeys) async {
+  if (pubkeys.isEmpty) return const [];
+  final entries = await Future.wait(
+    pubkeys.map((pubkey) async {
+      String? name;
+      try {
+        final metadata = await UserService(pubkey).getInitialMetadata();
```
pepina-dev: Now you are getting user metadata twice... first for sorting and then to display in the tile. Getting the metadata once per user should be enough, unless I am missing something here.

lib/screens/blocked_user_screen.dart
```diff
@@ -36,21 +41,97 @@ class BlockedUserScreen extends HookConsumerWidget {
...
+          systemNotice.showErrorNotice(context.l10n.failedToStartChat);
+        }
+      }
+    }
+
+    final isBlocked = blockState.isBlocked;
+
+    Widget bottomPanel;
+    if (isBlocked == false) {
+      bottomPanel = _UnblockedActionsPanel(
```
pepina-dev: This now looks like a repetition of the `StartChatScreen`. 

I think that would make more sense actually instead of repeating this logic is here is adding the blocked user system notice in that screen (maybe renaming it to UserProfileScreen).

mubarakcoded: @pepina-dev I get your point. If we move the block button into the existing screen (and rename the screen to UserProfileScreen), that would also make it visible on the start chat screen, too. Just wanted to confirm that this is the intended behavior. and gonna be lots of file changes, but I'm fine with that, in fact I'm almost through, just letting you know

pepina-dev: For me it's fine. @vladimir-krstic do you aggree?

ghost:
Ready to review this PR? Stage has broken it down into 7 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Add localization for blocked users UI](https://stagereview.app/marmot-protocol/whitenoise/pull/676/chapters/1) |
| 2 | [Refactor StartChatScreen into UserProfileScreen](https://stagereview.app/marmot-protocol/whitenoise/pull/676/chapters/2) |
| 3 | [Implement Blocked Users management screen](https://stagereview.app/marmot-protocol/whitenoise/pull/676/chapters/3) |
| 4 | [Hide chat composer for blocked peers](https://stagereview.app/marmot-protocol/whitenoise/pull/676/chapters/4) |
| 5 | [Update routing and deep links](https://stagereview.app/marmot-protocol/whitenoise/pull/676/chapters/5) |
| 6 | [Wire UI components to UserProfileScreen](https://stagereview.app/marmot-protocol/whitenoise/pull/676/chapters/6) |
| 7 | [Update changelog](https://stagereview.app/marmot-protocol/whitenoise/pull/676/chapters/7) |

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/676">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 53fec188bd5b63eb9dc725241b1563db7f9d9001 on May 21, 2026 2:56pm UTC.</sub>

lib/widgets/wn_system_notice.dart
```diff
@@ -231,6 +232,12 @@ class WnSystemNotice extends HookWidget {
           colors.backgroundContentPrimary,
           null,
         );
+      case WnSystemNoticeType.elevatedCard:
+        return (
+          colors.backgroundSlateElevated,
```
pepina-dev: I don't understand why this new system notice type is needed. From what I see in the figma, this system notice has in both themes consistent semantic color names: 

- background tertiary for the bg color
- background content primary for title color
- background content quaternary for description color

lib/routes.dart
```diff
@@ -398,14 +407,19 @@ abstract final class Routes {
           ),
         ),
         GoRoute(
-          name: 'startChat',
```
pepina-dev, line 401: My agent flagged this on another test, but it makes sense to me to change it now that the route does not exist:

> 🤖 The test helper router still registers `GoRoute(path: '/start-chat/:pubkey', …)`. Production code no longer routes there — `DeepLinks.parse` returns `/user-profile/<hex>` and the npub-tap path uses `UserProfileScreen.show` (Navigator push), not GoRouter. The assertion was rewritten to `find.byKey(ValueKey('user_route_$testPubkeyA')) … findsNothing` (line 982), which passes because nothing exercises that route any more.
> 
> Harmless today, but a NOOP that *looks* like a coverage signal. Future regressions in deep-link routing won't be caught here.
> 
> **Remediation:** Delete the `/start-chat/:pubkey` GoRoute. Replace the `findsNothing` assertion or remove it. If you want a router-level test that user deep links navigate to `/user-profile/:userPubkey`, add it explicitly (a related test exists in `test/routes_test.dart:198-203` — confirm it actually round-trips through `DeepLinks.parse`).

lib/screens/blocked_users_screen.dart
```diff
@@ -0,0 +1,158 @@
...
+    required this.onTap,
+  });
+
+  final String pubkey;
+  final FlutterMetadata? metadata;
+  final VoidCallback onTap;
+
+  @override
+  Widget build(BuildContext context) {
+    final displayName = presentName(metadata) ?? pubkey.substring(0, 8);
```
pepina-dev: My agent flagged this and I think it is right 🤖 


> When `presentName` returns null, the row shows the first 8 hex chars (e.g. `7f3b9c01…`). The npub line below at least starts with `npub1…`. Falling back to the truncated npub would match the rest of the codebase (`WnUserItem` usage).

pepina-dev (commented):
Another comment I noticed while testing and that it is an issue that is also in master but probably direct to address here as it is related to this touched screens: from what I see in the Figma, the user profile screen slate  (for blocked and unblocked users) should start from the top instead of being positioned at the bottom of the screen.

mubarakcoded:
> Another comment I noticed while testing and that it is an issue that is also in master but probably direct to address here as it is related to this touched screens: from what I see in the Figma, the user profile screen slate (for blocked and unblocked users) should start from the top instead of being positioned at the bottom of the screen.

Yes, that's the current behaviour, for blocked and unblocked users. It starts from the top of the screen

lib/screens/user_profile_screen.dart
```diff
@@ -233,7 +233,8 @@ class UserProfileScreen extends HookConsumerWidget {
           color: colors.backgroundContentSecondary,
         ),
       ),
-      type: WnSystemNoticeType.elevatedCard,
+      type: WnSystemNoticeType.neutral,
+      backgroundColor: colors.fillSecondary,
```
pepina-dev: Again, as I said in the comment before https://github.com/marmot-protocol/whitenoise/pull/676#discussion_r3278756647, the color in figma for background is background tertiary.
Image

lib/widgets/wn_system_notice.dart
```diff
@@ -39,6 +38,7 @@ class WnSystemNotice extends HookWidget {
     this.onToggle,
     this.autoHideDuration,
     this.animateEntrance = true,
+    this.backgroundColor,
```
pepina-dev: No, the variant should be enough for background color, I don't like this kind of overrides when the variant is supposed to define the style of the widget.

pepina-dev:
> Yes, that's the current behaviour, for blocked and unblocked users. It starts from the top of the screen

No, it is not. Tested in commit 85275b430

https://github.com/user-attachments/assets/dd18bdc1-3c7b-44b3-8f8f-24733758430d

mubarakcoded:
> No, it is not. Tested in commit [85275b4](https://github.com/marmot-protocol/whitenoise/commit/85275b430b3d27fdb2f2c795fafb05ab03449be1)

@pepina-dev Okay. I get your point, I thought you were referring to the view page from the block list, so our chat screen should be on top to align with the Figma design. fixing...

pepina-dev (changes requested):
As I mentioned before, it's in the bottom in the flow of start chat. Also, the blocked users notice has a padding that shouldn't be there 

|Figma| this PR|
|----|----|
|Image|Image

erskingardner (dismissed):

erskingardner (approved):

PR #679, opened 2026-05-19
Add 'Add members' button to group info

### LINKED ISSUE #643: [Feature]: Add member button in group info screen
The group info screen should have an 'add member' button to invite new members to the group directly from group settings.

dannym-arx:
marmot

## What

Admins can now add new members directly from the group info screen. Tap the new **Add members** button to open a picker, select followed users, and submit. The picker hides anyone who is already in the group.

## How

- New screen `add_group_members_screen.dart` reuses the `useUserSearch` and `useUserSelection` hooks from the new-group flow.
- New route `addGroupMembers` plus a `Routes.pushToAddGroupMembers` helper.
- Group info gets an admin-only button below **Edit group** that passes the current member list so the picker can filter them out.
- Submit calls `groups_api.addMembersToGroup` and pops back on success. Failures show an error notice and keep the user on the screen.

## Tests

- 8 widget tests for the new screen: header, candidate list, member filtering, button enable state, submit success, error notice, back navigation.
- 3 new tests in `group_info_screen_test.dart`: admin gating and navigation.
- `just precommit` passes.

A friendly marmot greets fellow marmots at the burrow door. That is the whole feature.

## Summary by CodeRabbit

* **New Features**
  * Administrators can now add members to groups using the new "Add members" button in the group info screen
  * New dedicated member addition interface with user search and filtering capabilities
  * Complete localization support added across 10 languages

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/679?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


dannym-arx:
@coderabbitai full review

lib/routes.dart
```diff
@@ -679,6 +695,18 @@ abstract final class Routes {
     );
   }
 
+  static void pushToAddGroupMembers(
```
erskingardner: > The member list snapshot at the moment the button is tapped is serialized into the route's `extra` field and then deserialized on the other side:
> 
> ```dart
> final existingMembers = state.extra as List<String>? ?? const [];
> ```
> 
> This works correctly for the happy path, but `state.extra` survives deep linking and process restoration only when it is a serializable type that go_router can handle. For `List<String>` go_router does not re-serialize `extra` when restoring navigation state after the app is killed and resumed. The member list will come back as `null`, and the filter falls back to `const []`, so all current members will appear as candidates. A user who was added before the app was killed could be added again via this screen.
> 
> A safer approach is to fetch members inside `AddGroupMembersScreen` itself (or in a dedicated hook) using the `groupId` that is already in the path parameter, rather than relying on `extra`. The new screen already has `groupId` — it just needs to call `groups_api.groupMembers` on load and apply the filter from live data.

lib/screens/add_group_members_screen.dart
erskingardner, line 1: is the UserSelectionScreen going to go away? Should we merge these a bit more? 

> The 224-line new screen shares ~180 lines of identical structure with `UserSelectionScreen` (`WnSlate` + search field + bubble row + empty state + `ListView.builder` + fade overlays). The only differences are the submit action and the `candidates` filter. This is a strong signal that `UserSelectionScreen` should be made more general, or a shared `UserPickerScreen`/widget should be extracted, rather than maintaining two near-identical screens. At minimum, the duplication makes future visual changes (e.g., bubble height, overlay colors, loading indicator style) require two edits instead of one.

lib/routes.dart
dannym-arx: Good catch, fixed in daea87b. The screen now uses `useGroupMembers` with the `groupId` from the route, so the filter works regardless of how the screen is entered. Added a test for the members-fetch failure path too.

lib/screens/add_group_members_screen.dart
```diff
@@ -0,0 +1,70 @@
...
+    return UserPickerScreen(
+      title: context.l10n.addMembers,
+      submitText: context.l10n.addMembers,
+      submitIcon: WnIcons.userFollow,
+      submitErrorMessage: context.l10n.failedToAddMembers,
+      additionalLoading: membersState.isLoading,
+      additionalNotice: fetchNotice,
+      candidateFilter: (user) => !existingMembersSet.contains(user.pubkey),
+      onSubmit: (ctx, selected) async {
+        await groups_api.addMembersToGroup(
```
pepina-dev: Why is this re implemented instead of using the method to add members from `useGroupMembers` hook? It has a method called `addMembers`

lib/screens/add_group_members_screen.dart
```diff
@@ -0,0 +1,70 @@
...
+    final membersState = useGroupMembers(
+      accountPubkey: accountPubkey,
+      groupId: groupId,
+    );
+    final fetchErrorMessage = useState<String?>(null);
+    final fetchErrorLabel = context.l10n.failedToFetchGroupMembers;
+
+    useEffect(() {
+      if (membersState.error != null) {
+        fetchErrorMessage.value = fetchErrorLabel;
```
pepina-dev: This error is a key that could be something different that failed to fetch members (if add is used from hook instead of duplicated)  but will always show this label 🤔

lib/screens/user_picker_screen.dart
```diff
@@ -0,0 +1,229 @@
+import 'package:flutter/material.dart';
```
pepina-dev, line 1: If this is a screen, why it doesn't live in screens folder? Naming here is confusing. UserSelectionScreen and UserPickerScreen sound like the same to me, is not evident that the UserSelectionScreen  actually imports this one.

erskingardner (approved):

PR #686, opened 2026-05-20
fix: route user-tag taps to the group member screen for members

### LINKED ISSUE #681: User tag navigation
# Context
When clicking in a user tag in a message bubble, it shows the start chat screen. This is a bit unexpected comparing with other chat apps (like whatsapp or signal) where tags navigate to the same screen that you woudl see when clicking on a group member from the members lists screen.The reason for doing this is that tags in messages may tag users that are not in the group. 

# Solution
A solution that we discussed here https://github.com/marmot-protocol/whitenoise/pull/665#discussion_r3261928955 would be to keep the start chat navigation only for users that are not members of the group and for group members navigate to the group member screen

⚠️ We don't want to fetch group members in every chat bubble

dannym-arx:
marmot

Closes #681.

Tapping an @-mention inside a chat bubble used to open the start-chat shade, even when the tagged marmot was already in the group. Now the bubble checks the group roster and routes to the group member shade for members, keeping the start-chat shade for non-members and DMs.

Since the chat screen already loads the member list once via `useGroupMembers`, the bubble gets a memoized `Set<String>` instead of fetching members per bubble. That keeps the "don't fetch in every chat bubble" line from the issue satisfied.

For consistency, the start-chat shade now anchors to the top of the screen, matching the group member shade. Both shades slide in from the same spot.

https://github.com/user-attachments/assets/5162955e-a366-46ab-bfc2-ee186768fc24

## Summary by CodeRabbit

* **New Features**
  * Tapping a user mention in group chat opens the group-member profile when available; if the user isn’t a group member it opens a direct chat.

* **Style**
  * Adjusted start chat screen spacing for improved layout balance.

* **Tests**
  * Added tests for mention navigation and non-member redirect behavior to validate routing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/marmot-protocol/whitenoise/pull/686?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/screens/chat_screen.dart
+++ b/lib/screens/chat_screen.dart
@@ -207,6 +207,10 @@ class ChatScreen extends HookConsumerWidget {
       },
       [mentionMembers],
     );
+    final groupMemberPubkeySet = useMemoized(
+      () => mentionMemberPubkeys.toSet(),
+      [mentionMemberPubkeyKey],
+    );
     String? resolveMentionDisplayName(String hexPubkey) =>
         mentionNamesByPubkey[hexPubkey] ?? presentName(getAuthorMetadata(hexPubkey));
     final resolveMentionDisplayNameRef = useRef(resolveMentionDisplayName);
@@ -547,6 +551,8 @@ class ChatScreen extends HookConsumerWidget {
                   : null,
               onHorizontalDragEnd: isSearchMode ? null : () => input.setReplyingTo(message),
               mentionDisplayName: resolveMentionDisplayName,
+              groupId: groupId,
+              groupMemberPubkeys: isGroupChat ? groupMemberPubkeySet : null,
               onRetry:
                   !isSearchMode && isOwnMessage && message.deliveryStatus is DeliveryStatus_Failed
                   ? () async {
--- a/lib/screens/chat_screen.dart
+++ b/lib/screens/chat_screen.dart
@@ -207,10 +207,6 @@ class ChatScreen extends HookConsumerWidget {
       },
       [mentionMembers],
     );
-    final groupMemberPubkeySet = useMemoized(
-      () => mentionMemberPubkeys.toSet(),
-      [mentionMemberPubkeyKey],
-    );
     String? resolveMentionDisplayName(String hexPubkey) =>
         mentionNamesByPubkey[hexPubkey] ?? presentName(getAuthorMetadata(hexPubkey));
     final resolveMentionDisplayNameRef = useRef(resolveMentionDisplayName);
@@ -551,8 +547,7 @@ class ChatScreen extends HookConsumerWidget {
                   : null,
               onHorizontalDragEnd: isSearchMode ? null : () => input.setReplyingTo(message),
               mentionDisplayName: resolveMentionDisplayName,
-              groupId: groupId,
-              groupMemberPubkeys: isGroupChat ? groupMemberPubkeySet : null,
+              groupId: isGroupChat ? groupId : null,
               onRetry:
                   !isSearchMode && isOwnMessage && message.deliveryStatus is DeliveryStatus_Failed
                   ? () async {
--- a/lib/screens/group_member_screen.dart
+++ b/lib/screens/group_member_screen.dart
@@ -95,11 +95,10 @@ class GroupMemberScreen extends HookConsumerWidget {
       if (membersState.members.contains(memberPubkey)) return null;
       didRedirect.value = true;
       WidgetsBinding.instance.addPostFrameCallback((_) {
-        final nav = Routes.navigatorKey.currentContext;
-        if (nav == null || !nav.mounted) return;
-        Routes.goBack(nav);
-        if (!nav.mounted) return;
-        StartChatScreen.show(nav, userPubkey: memberPubkey);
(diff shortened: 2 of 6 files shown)


lib/screens/chat_screen.dart
```diff
@@ -207,6 +207,10 @@ class ChatScreen extends HookConsumerWidget {
       },
       [mentionMembers],
     );
+    final groupMemberPubkeySet = useMemoized(
```
pepina-dev: Group members are loaded only on mount. If a user is added while I'm in the chat, when clicking its mention it redirects to start chat although it is a member. What I'm thinking is that maybe the group member screen could do this check, instead of trusting the navigation. And If it's not member of the group, then redirect to the start chat?

dannym-arx: Good catch, pushed cadbf99d. The bubble no longer caches a member set; it always navigates to the group member screen for group chats, and the screen itself checks fresh membership via its own `useGroupMembers` load. If the tagged user is not in the loaded list, it redirects to the start-chat shade. So a user added mid-session now resolves correctly.

lib/screens/group_member_screen.dart
```diff
@@ -87,6 +88,22 @@ class GroupMemberScreen extends HookConsumerWidget {
...
     }, [membersState.error]);
 
+    final didRedirect = useRef(false);
+    useEffect(() {
+      if (didRedirect.value) return null;
+      if (membersState.isLoading || membersState.error != null) return null;
+      if (membersState.members.contains(memberPubkey)) return null;
+      didRedirect.value = true;
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        final nav = Routes.navigatorKey.currentContext;
```
pepina-dev: My agent flagged this now 🤖 . I hadn't seen this Routes.navigatorKey.currentContext before, so it made more sense to me to use context.mounted instead ... WDYT?

> ####  Use the widget's own `BuildContext`, not `Routes.navigatorKey.currentContext`
> **Where:** `lib/screens/group_member_screen.dart:97-103`
> 
> Today:
> ```dart
> final nav = Routes.navigatorKey.currentContext;
> if (nav == null || !nav.mounted) return;
> Routes.goBack(nav);
> if (!nav.mounted) return;
> StartChatScreen.show(nav, userPubkey: memberPubkey);
> ```
> 
> The widget already has a `BuildContext` in scope — it's the `context` parameter on `build`, which the `useEffect` closure captures synchronously. The post-frame callback can use that directly. Reaching through `Routes.navigatorKey.currentContext` to recover a context the widget already owns:
> 
> - Adds an unnecessary cross-cutting dependency on `Routes.navigatorKey` from a screen file that otherwise has no business knowing about the global navigator key.
> - Forces two `mounted` guards because the resolved context is now untrusted by construction; one `context.mounted` guard on the widget's own context would do.
> - Makes the redirect's correctness depend on the screen still being mounted under the *root* navigator. Today that's always true (`Routes.navigatorKey` IS the GoRouter's navigator — `lib/routes.dart:133`), but if anyone later introduces a `ShellRoute` or `StatefulShellRoute`, the global key stops pointing at the navigator that actually contains `GroupMemberScreen`. The widget-context version is immune to that change because the context resolves whatever navigator the widget is mounted under.
> 
> **Proposed change:**
> ```dart
> WidgetsBinding.instance.addPostFrameCallback((_) {
>   if (!context.mounted) return;
>   Routes.goBack(context);
>   if (!context.mounted) return;
>   StartChatScreen.show(context, userPubkey: memberPubkey);
> });
> ```
> 
> `Routes.goBack` (`lib/routes.dart:514`) and `StartChatScreen.show` both take a `BuildContext` already — this just uses them the way they're meant to be used. Drop the `Routes.navigatorKey` reference from this file entirely; nothing else here needs it.

pepina-dev (commented):
One last comment 😬 🤖

ghost:
Ready to review this PR? Stage has broken it down into 5 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Adjust StartChatScreen layout and anchoring](https://stagereview.app/marmot-protocol/whitenoise/pull/686/chapters/1) |
| 2 | [Update ChatMessageBubble to support group routing](https://stagereview.app/marmot-protocol/whitenoise/pull/686/chapters/2) |
| 3 | [Pass groupId from ChatScreen to message bubbles](https://stagereview.app/marmot-protocol/whitenoise/pull/686/chapters/3) |
| 4 | [Implement non-member redirect in GroupMemberScreen](https://stagereview.app/marmot-protocol/whitenoise/pull/686/chapters/4) |
| 5 | [Verify mention navigation and redirection with tests](https://stagereview.app/marmot-protocol/whitenoise/pull/686/chapters/5) |

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/686">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 3baebc40fbe0ac6e13b1b7b3b14d2020cc31dcac on May 20, 2026 7:56pm UTC.</sub>

pepina-dev (approved):
LGTM

PR #701, opened 2026-06-15
fix(ios): stop 0xdead10cc account-creation crash and fix lost-login error message

vincenzopalazzo:
## Summary

Two linked iOS fixes for the "can't create the account / White Noise Crashed" report.

- **fix(ios): relax App Group data protection** — stops the `0xdead10cc` SIGKILL that crashes the app (and the NSE) during account creation/login.
- **fix(login): accurate error on the Relay Setup screen** — a lost pending login now shows "No login in progress. Please start over." instead of the misleading generic "An error occurred during login. Please try again."

### The crash (`0xdead10cc`)

Pulled the real crash reports off-device (`idevicecrashreport`, no reinstall). Both processes were terminated by RUNNINGBOARD:

```
Runner-2026-06-12.ips                                 EXC_CRASH (SIGKILL)
WhiteNoiseNotificationServiceExtension-2026-06-12.ips EXC_CRASH (SIGKILL)
termination: RUNNINGBOARD, code 3735883980 = 0xDEAD10CC   (build 2026.5.22+25)
```

`0xdead10cc` ("deadlock") is iOS killing a process that **holds a SQLite/file lock on a file in a shared container across suspension**. The chain here:

- The encrypted DB lives in the **App Group** container (`group…/whitenoise/data`) so the NSE can read it.
- The core opens it in **WAL mode with a persistent pool** — it holds `-wal`/`-shm` locks.
- **No file sets a data-protection class**, so files default to `NSFileProtectionComplete`, which is unavailable while the device is locked.
- During account creation/login the relay step does slow network work; if the screen locks or the app is suspended while the WAL lock is held → `0xdead10cc`.

Fix: apply `NSFileProtectionCompleteUntilFirstUserAuthentication` to the `data`/`logs` dirs (and, once, to files that predate the fix) in both `Runner` and the NSE, before the DB is opened. New files inherit the directory's class, so the recursive downgrade is gated by a marker (`.wn_data_protection_v1`) to avoid walking the media cache on every launch / every push. This also lets the NSE read the DB while the device is locked (which it must, to render push content).

### The login banner (image with "An error occurred during login")

This is the same incident's second symptom. The pending login is an in-memory `DashMap` (`pending_logins`, `AccountManager`); it is wiped when the process is killed (the crash above) or the core reinitializes. After that, the relay-publish step returns `LoginError::NoLoginInProgress`.

The Relay Setup screen's error map (`use_relay_resolution.dart`) only handled `NoRelayConnections`/`Timeout`/`Internal`, so `NoLoginInProgress` fell through to the generic retryable message — but retrying there can never succeed; the user must restart from key entry. Tracing the core confirms `NoLoginInProgress` is the only path to the generic message at that screen (keychain and relay failures all map to specific, already-handled errors).

Fix: mirror the login screen's map on the relay screen (`NoLoginInProgress`, `InvalidKeyFormat`) using existing l10n strings — no new translation keys. `NoLoginInProgress` now renders "No login in progress. Please start over."

## Test plan

- `flutter analyze` — clean.
- `flutter test` — full suite green (4736 tests), including new cases:
  - `main_test.dart`: `applyIosDataProtection` invokes the channel on iOS, is a no-op elsewhere, swallows platform exceptions.
  - `use_relay_resolution_test.dart` + `relay_resolution_screen_test.dart`: `NoLoginInProgress` maps to the "start over" message.
- Swift helper type-checked against the iOS simulator SDK.

## Verification note

The data-protection change can't be fully verified at runtime in this environment: file data protection is a no-op on the Simulator, and reinstalling on a device wipes the App Group DB. Recommend on-device verification via `xcrun devicectl device install` over the existing install (do not `flutter install`, which may uninstall first). The root robustness of the in-memory `pending_logins` lives in `whitenoise-rs`, out of scope for this app-layer PR; this PR makes the symptom accurate and removes the crash that wipes the session.

---

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/701">
  <picture>
    Open in Stage
  </picture>
</a>

## Release Notes

* **New Features**
  * Enhanced iOS data protection for the app’s stored `data` and `logs`, including a one-time safeguard for existing files.
  * Added iOS startup support for applying data protection to multiple directories via a platform method call.
* **Bug Fixes**
  * Improved relay-resolution error handling with a new “No login in progress” message and matching localized UI text.
* **Tests**
  * Added coverage for the new login error mapping and iOS data-protection behavior during startup.

### DIFF
--- a/ios/Runner/AppDelegate.swift
+++ b/ios/Runner/AppDelegate.swift
@@ -92,6 +92,8 @@ private func wn_install_ios_background_keyring_store_with_access_group(
       switch call.method {
       case "getAppGroupContainerPath":
         result(self.appGroupContainerPath())
+      case "applyDataProtection":
+        result(self.applyDataProtection(arguments: call.arguments))
       default:
         result(FlutterMethodNotImplemented)
       }
@@ -189,6 +191,16 @@ private func wn_install_ios_background_keyring_store_with_access_group(
     trigger == "invite" || trigger == "group_invite" || trigger == "GroupInvite"
   }
 
+  private func applyDataProtection(arguments: Any?) -> Bool {
+    guard
+      let args = arguments as? [String: Any],
+      let paths = args["paths"] as? [String]
+    else {
+      return false
+    }
+    return paths.allSatisfy { WhiteNoiseDataProtection.apply(atPath: $0) }
+  }
+
   private func appGroupContainerPath() -> String? {
     guard
       let identifier = Bundle.main.object(forInfoDictionaryKey: "WNAppGroupIdentifier") as? String,
@@ -256,6 +268,62 @@ private func wn_install_ios_background_keyring_store_with_access_group(
     let bytes = Array(string.utf8)
     return bytes.withUnsafeBufferPointer { buffer in
       body(buffer.baseAddress, buffer.count)
+    }
+  }
+}
+
+/// Applies `completeUntilFirstUserAuthentication` data protection to a directory
+/// and everything inside it.
+///
+/// The White Noise database lives in the shared App Group container so the
+/// notification service extension can read it. Files there default to a
+/// protection class that is unavailable while the device is locked; holding a
+/// SQLite/WAL lock on such a file in a shared container across suspension makes
+/// iOS terminate the process with 0xdead10cc. Setting the relaxed class keeps
+/// the files reachable after first unlock and avoids that termination.
+enum WhiteNoiseDataProtection {
+  private static let markerName = ".wn_data_protection_v1"
+
+  @discardableResult
+  static func apply(atPath path: String) -> Bool {
+    let fileManager = FileManager.default
+    let attributes: [FileAttributeKey: Any] = [
+      .protectionKey: FileProtectionType.completeUntilFirstUserAuthentication
+    ]
+    // Always set the class on the directory itself; new files (db, -wal, -shm,
+    // media cache) inherit it, so this is the only step needed on most launches.
+    var success = setAttributes(attributes, atPath: path, using: fileManager)
+
+    // Downgrade files that predate this fix exactly once. Recursing the whole
+    // data dir (which holds the media cache) on every launch would be unbounded
+    // work, so a marker gates it after the first successful pass.
+    let markerPath = (path as NSString).appendingPathComponent(markerName)
+    guard !fileManager.fileExists(atPath: markerPath) else {
+      return success
+    }
+    if let enumerator = fileManager.enumerator(atPath: path) {
+      for case let relativePath as String in enumerator {
+        let itemPath = (path as NSString).appendingPathComponent(relativePath)
+        success = setAttributes(attributes, atPath: itemPath, using: fileManager) && success
+      }
+    }
+    if success {
+      fileManager.createFile(atPath: markerPath, contents: nil, attributes: attributes)
+    }
+    return success
+  }
+
+  private static func setAttributes(
+    _ attributes: [FileAttributeKey: Any],
+    atPath path: String,
+    using fileManager: FileManager
+  ) -> Bool {
+    do {
+      try fileManager.setAttributes(attributes, ofItemAtPath: path)
+      return true
+    } catch {
+      NSLog("White Noise failed to set data protection on %@: %@", path, error.localizedDescription)
+      return false
     }
   }
 }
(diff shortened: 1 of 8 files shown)


ghost:
Ready to review this PR? Stage has broken it down into 4 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Implement iOS data protection utility](https://stagereview.app/marmot-protocol/whitenoise/pull/701/chapters/1) |
| 2 | [Wire data protection to iOS app and extension](https://stagereview.app/marmot-protocol/whitenoise/pull/701/chapters/2) |
| 3 | [Invoke iOS data protection from Flutter](https://stagereview.app/marmot-protocol/whitenoise/pull/701/chapters/3) |
| 4 | [Fix login error messaging on Relay screen](https://stagereview.app/marmot-protocol/whitenoise/pull/701/chapters/4) |

<a href="https://stagereview.app/marmot-protocol/whitenoise/pull/701">
  <picture>
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 2ee9e10206f8065bdc63646cb66639848d9f842e on Jun 15, 2026 12:06pm UTC.</sub>

vincenzopalazzo (commented):
Self-review: one issue to fix before merge, plus one non-blocking note. The inline specifics are below.

## Code Review

### Positive aspects
- Root cause is backed by real device crash reports (`0xdead10cc` RUNNINGBOARD SIGKILL on both `Runner` and the NSE), not a guess.
- Setting the protection class on the *directory* and gating the recursive downgrade behind a marker (`.wn_data_protection_v1`) is the right call — new `-wal`/`-shm`/media files inherit the class, so neither startup nor the time-boxed NSE walks the media cache on every run.
- The error-map fix reuses existing l10n keys (no new translations), and traces cleanly to the single reachable cause (`LoginError::NoLoginInProgress`).

---

### Issues & suggestions

#### 1. Drop the unreachable `LoginInvalidKeyFormat` arm on the relay screen (Medium)
`lib/hooks/use_relay_resolution.dart:45` and `lib/screens/relay_resolution_screen.dart` add a `LoginInvalidKeyFormat -> loginErrorInvalidKey` mapping. That error can't reach this screen: the key is validated in `login_start`, and the relay-publish path (`login_publish_default_relays` / `login_with_custom_relay`) never re-parses it. So the arm is dead, and it's **untested** — the repo's 99% coverage gate (`just coverage`) will fail on the new branch. Keep only `LoginNoLoginInProgress` (the actual bug) and drop `LoginInvalidKeyFormat` from both the hook map and `_resolveError`.

#### 2. Concurrent one-time recursion in app + NSE (Cleanup, non-blocking)
If a push fires while the app is launching, both processes can run the one-time recursive downgrade before either writes the marker. It's idempotent (`setAttributes` to the same class), so harmless — just noting it's a known, benign race, not a correctness issue.

---

### Summary
LGTM after dropping the unreachable `LoginInvalidKeyFormat` arm (item 1) so coverage stays green. Item 2 is non-blocking.

---

### Decision Log

**Hardest decision:** Whether to fix `0xdead10cc` at the file-protection layer or by closing DB connections on background. File protection (`completeUntilFirstUserAuthentication`) is the minimal, well-trodden fix and also satisfies the NSE's need to read the DB while the device is locked; forcing connection teardown would fight the core's persistent sqlx pool in `whitenoise-rs`, which this app layer doesn't own.

**Alternatives rejected:**
- Recurse the data dir on every launch/push: simpler, but unbounded main-thread/NSE work over the media cache — rejected for the marker-gated one-time pass.
- Persist `pending_logins` so login survives a restart: the right deep fix, but it lives in `whitenoise-rs`, out of scope for this app PR. This PR makes the symptom accurate and removes the crash that wipes the session.
- Mirror the login screen's full error map verbatim: led to the dead `InvalidKeyFormat` arm in item 1 — scope-trimmed to what's reachable.

**Least confident about:** The runtime behavior of the protection change can't be verified here — data protection is a no-op on the Simulator and a device reinstall wipes the App Group DB. I'm relying on the documented `0xdead10cc` semantics and the crash-log evidence; on-device confirmation via `xcrun devicectl device install` over the existing install is the real test.

vincenzopalazzo:
Addressed the self-review.

**Item 1 (Medium) — fixed in 2ee9e10.** Dropped the unreachable `LoginInvalidKeyFormat` arm from both `use_relay_resolution.dart` and `_resolveError`; the key is validated in `login_start` and the relay-publish path never re-parses it, so the branch was dead and untested (would have tripped the 99% `just coverage` gate). The relay screen now maps only `LoginNoLoginInProgress` -> "No login in progress. Please start over." Folded into the login commit to keep the two-commit history clean rather than stacking a fixup.

**Item 2 (Cleanup) — non-blocking, left as-is.** The app/NSE one-time recursion race is idempotent (`setAttributes` to the same class), so no change.

Local checks on the new head: `flutter analyze` clean, `flutter test` green (relay hook + screen suites). Full Rust/Android CI is gated on the fork PR pending a maintainer run.

